PR #1, opened 2021-08-13
Repo Setup and Aborting Panic Handler

jcape:
This PR creates a simple abort-on-panic crate which can be included into an SGX enclave that's already linking to `libsgx_trts.a`, and compiled with a panic strategy of abort.

### DIFF
--- a/panic/abort/README.md
+++ b/panic/abort/README.md
@@ -10,7 +10,7 @@ Aborting panic handler for Intel-SGX enclaves.
 [![Crates.io Downloads][downloads-image]][crate-link]
 [![Build Status][build-image]][build-link]
 
-This crate provides a panic handler implementation intended to be executed inside an SGX enclave. Specifically, it will provide the required `rust_eh_personality()` and `#![panic_handler]` methods to satisfy the linker for a no-std environment.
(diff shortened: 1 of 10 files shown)


eranrund:
Nice. Looking forward to seeing the implementation.
The GH actions look call, looking forward to seeing how that works as well.

eranrund (approved):

panic/abort/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+[![crate][crate-image]][crate-link]
+[![Docs][docs-image]][docs-link]
+![GPLv3.0 Licensed][license-image]
+![Rust Version][rustc-image]
+[![Project Chat][chat-image]][chat-link]
+[![Crates.io Downloads][downloads-image]][crate-link]
+[![Build Status][build-image]][build-link]
+
+This crate provides a panic handler implementation intended to be executed inside an SGX enclave. Specifically, it will provide the required `rust_eh_personality()` and `#![panic_handler]` methods to satisfy the linker for a `#![no_std]` environment. As a result, this crate will
```
mfaulk, line 13: "As a result, this crate will" ...?

christian-oudard (approved):
Looks good, pending the documentation fix mentioned by mfaulk.

garbageslam:
I think this is a big regression, we're not logging the panics the way the current `mc-sgx-panic` crate is. I think we need to revisit this and bring back that functionality.

PR #2, opened 2022-05-05
Add rust interface to untrusted SGX

nick-mobilecoin:
Add a rust interface for the untrusted side of the SGX interface.  This
provides the capability to create enclaves and call into them.

This is bigger than I wanted it to be, but I wanted to have working tests that show the rust wrapper actually works.
Up for suggestions on how this can be broken up more.

This place the untrusted  interface down in a lower directory as its own cargo workspace, this to let it use the host target triple while the upcoming trusted side will utilize a custom target triple.

This was not played on top of the `develop` branch to minimize the already significant changes


untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,253 @@
...
+        .arg(dynamic_enclave.to_str().expect("Invalid UTF-8 in static enclave path"))
+        .args(&["-z", "relro", "-z", "now", "-z", "noexecstack"])
+        .arg(&format!("-L{}/lib64/cve_2020_0551_load", sgx_library_path()))
+        .arg(&format!("-L{}/lib64", sgx_library_path()))
+        .arg("--no-undefined")
+        .arg("--nostdlib")
+        .arg("--start-group")
+        .args(&["--whole-archive", "-lsgx_trts_sim", "--no-whole-archive"])
+        .arg(static_enclave.as_ref().to_str().unwrap())
+        .args(&["-lsgx_tstdc", "-lsgx_tcxx", "-lsgx_tcrypto", "-lsgx_tservice_sim"])
```
remoun, line 155: Do we always want the `_sim` libraries here?

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,253 @@
...
+        .arg("--end-group")
+        .arg("-Bstatic")
+        .arg("-Bsymbolic")
+        .arg("--no-undefined")
```
remoun, line 159: Nit: This arg was already passed on line 150

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,253 @@
...
+        .arg("-eenclave_entry")
+        .arg("--export-dynamic")
+        .args(&["--defsym", "__ImageBase=0"])
+        .arg("--gc-sections")
+        .arg(&format!("--version-script={}", ENCLAVE_LINKER_SCRIPT));
+
+    let status = command.status().expect("Failed to run the linker for dynamic enclave");
+    match status.code().unwrap() {
+        0 => (),
+        _ => panic!("Failed to link the dynamic enclave")
```
remoun, line 170:
```suggestion
        code => panic!("Linker exited with code {}", code)
```

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+    dynamic_enclave
+
```
remoun, line 173: Nit: Unnecessary blank line
```suggestion
```

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,253 @@
...
+/// ready for use in `sgx_create_enclave()`.
+fn sign_enclave_binary<P: AsRef<Path>>(unsigned_enclave: P) -> PathBuf {
+    let mut signed_binary = PathBuf::from(unsigned_enclave.as_ref());
+    signed_binary.set_extension("signed.so");
+
+    let mut command = Command::new(format!("{}/bin/x64/sgx_sign", sgx_library_path()));
+    command.arg("sign") .arg("-enclave") .arg(unsigned_enclave.as_ref())
+        .arg("-config") .arg(ENCLAVE_CONFIG)
+        .arg("-key") .arg(SIGNING_KEY)
+        .arg("-out") .arg(&signed_binary);
```
remoun, line 193: Nit: Move calls one per line
```suggestion
    command
        .arg("sign")
        .arg("-enclave")
        .arg(unsigned_enclave.as_ref())
        .arg("-config")
        .arg(ENCLAVE_CONFIG)
        .arg("-key")
        .arg(SIGNING_KEY)
        .arg("-out")
        .arg(&signed_binary);
```

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,253 @@
...
+    let status = command.status().expect("Failed to execute enclave signer");
+    match status.code().unwrap() {
+        0 => (),
+        _ => panic!("Failed to sign enclave")
```
remoun, line 197:
```suggestion
        code => panic!("sgx_sign exited with code {}", code)
```

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,253 @@
...
+/// partially a build test.
+///
+/// # Arguments
+///
+/// * `untrusted_file` - The untrusted C file generated from edger8
+///
+/// # Returns
+/// The full path to resultant untrusted library.
+fn build_untrusted_library<P: AsRef<Path>>(untrusted_file: P) -> PathBuf {
+
```
remoun, line 216: Nit: Unnecessary blank line
```suggestion
```

untrusted/test_enclave/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+
+/// The location to the test enclave.  This is determined at compile time.
+/// As such moving the enclave binary after the fact will *not* update this
+/// value.
+pub static ENCLAVE: &str = concat!(env!("OUT_DIR"), "/libenclave.signed.so");
```
remoun, line 9: How about naming this ENCLAVE_PATH?
```suggestion
pub static ENCLAVE_PATH: &str = concat!(env!("OUT_DIR"), "/libenclave.signed.so");
```

untrusted/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    /// # Returns
+    ///
+    /// [_status_t_SGX_SUCCESS] when the enclave is created successfully.
+    ///
+    /// See
+    /// <https://download.01.org/intel-sgx/sgx-dcap/1.13/linux/docs/Intel_SGX_Enclave_Common_Loader_API_Reference.pdf>
+    /// for error codes and their meaning.
+    pub fn create(&mut self) -> sgx_status_t {
+        let mut launch_token: sgx_launch_token_t = [0; 1024];
```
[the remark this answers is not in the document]
nick-mobilecoin: good catch!

Seems like I should update the doc comment to say this function should only every be called once per instance. 
I can't think of a term for a function like that, "re-entrant" is close but slightly different as it usually implies calling the function more than once concurrently.

While calling this function more than once is a slightly different error case than failing to create the enclave from the C API, I'm not sure the rust API should have different ways to fail, that seems inconsistent for an eventual  end user.
Perhaps for now this should return something like `_status_t_SGX_ERROR_UNEXPECTED` (I don't like that error code, but...).   Then longer term or as a follow on this should return a `Result` such that `_status_t_SGX_SUCCESS` => Ok(). and the other values come back as an `Error` enum.  Then an error like `InUse` could be returned for this attempted recreation logic.

untrusted/test_enclave/src/lib.rs
nick-mobilecoin: renamed as suggested

untrusted/test_enclave/build.rs
nick-mobilecoin: I have no problem opening the tests up to HW if we think it will be easy to utilize in CI.  If we end up not having a constant HW run going, I would be concerned that it breaks too easily.

I still have access to VM with HW sgx if we want to pursue that now

remoun: It seems the GHA runners, both generic and self-hosted, support SGX. So if it's not too much extra work, supporting HW would be nice. It can also remain a TODO until we need it.

untrusted/src/lib.rs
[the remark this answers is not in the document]
remoun: One way to implement this is with a Builder pattern. Then you can have a method that consumes the instance, e.g. `fn create(self) -> Enclave`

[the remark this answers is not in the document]
nick-mobilecoin: I don't think I fully follow. Are you saying to return itself in the result
```
impl Enclave {
    pub fn create(self) -> Result<Enclave, Error>
```
or are you thinking of having 2 types
```
impl Builder {
    pub fn create(self) -> Result<Enclave, Error>

File: untrusted/build.rs [hunk 10]
```diff
@@ -0,0 +1,54 @@
...
+        mode => {
+            warning!(
+                "'SGX_MODE' was set to '{}'. Should be one of 'SW' or 'HW', defaulting to 'SW'",
+                mode
+            );
+            "_sim"
+        }
+    };
+    String::from(suffix)
+}
```
jcape, line 28: This should be decided by a "sim" feature.

untrusted/src/lib.rs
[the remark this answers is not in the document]
remoun: The latter

untrusted/src/lib.rs
```diff
@@ -0,0 +1,157 @@
+// Copyright (c) 2022 The MobileCoin Foundation
```
jcape, line 1: The first crate should handle the linkage and bindings and export the API (i.e. a pure FFI crate).

The *second* crate would be the one that has the ideomatic rust types (this also means you can nuke various "allow(thing_that_looks_like_c)" stuff.

untrusted/src/lib.rs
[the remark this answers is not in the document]
jcape: Yeah, in both cases, all the `Enclave::create()` code should go into `Enclave::new()`, the only difference is whether `Enclave::new()` is public or not.

untrusted/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    /// * `filename` - The name of the enclave file.  This should be a signed
+    ///     enclave.
+    pub fn new(filename: &str) -> Enclave {
+        let filename = CString::new(filename).expect("Can't convert enclave filename to CString.");
+        Enclave {
+            filename,
+            ..Default::default()
+        }
+    }
```
jcape, line 38: Few things:
1. We should be creating enclaves from bytes, not files, so we can bake them into the app via `include_bytes!()`. So this should take a byte slice (`&[u8]`) instead.
2. When doing file-based APIs, you should use `&Path` instead of `&str`, and `PathBuf` instead of `String`.
3. As noted below, this should be fallible and include the contents of `create()`

untrusted/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+        let mut enclave_id: sgx_enclave_id_t = 0;
+        let result = unsafe {
+            sgx_create_enclave(
+                self.filename.as_ptr(),
+                self.debug as c_int,
+                &mut launch_token as *mut sgx_launch_token_t,
+                &mut launch_token_updated,
+                &mut enclave_id,
+                &mut misc_attr,
+            )
```
jcape, line 88: As noted above, this should use [`sgx_create_enclave_from_buffer_ex`](https://download.01.org/intel-sgx/sgx-linux/2.16/docs/Intel_SGX_Developer_Reference_Linux_2.16_Open_Source.pdf#%5B%7B%22num%22%3A232%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C339%2C0%5D)

untrusted/test_enclave/build.rs
jcape: We need to be able to run `cargo test` on desktops, which don't support SGX, so we need to keep the sim versions of these accessible.

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,258 @@
+// Copyright (c) 2022 The MobileCoin Foundation
```
jcape, line 1: We already have code that does everything in this build script in the main repo. If we need to refactor it out into this repo (or another repo), we should just do that.

untrusted/test_enclave/src/config.xml
```diff
@@ -0,0 +1,15 @@
...
+    <TCSMinPool>1</TCSMinPool>
+    <TCSPolicy>1</TCSPolicy>
+    <StackMaxSize>0x40000</StackMaxSize>
+    <HeapMaxSize>0x100000</HeapMaxSize>
+    <HeapInitSize>0x100000</HeapInitSize>
+    <HeapMinSize>0x100000</HeapMinSize>
+    <DisableDebug>0</DisableDebug>
+    <MiscSelect>0</MiscSelect>
+    <MiscMask>0xFFFFFFFF</MiscMask>
+</EnclaveConfiguration>
```
jcape, line 15: As noted, there is code that handles constructing this file in a repeatable way in the repo, we should use that code.

untrusted/test_enclave/src/enclave.c
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+/*
+ * A basic function that just adds 2 to its input
+ */
+void ecall_add_2(int input, int *sum) {
+    *sum = input + 2;
+}
```
jcape, line 7: This should be in rust.

```
#[no_mangle]
pub fn ecall_add_2(input: i32, sum: &i32) {
    *sum = input + 2;
}
```
should do the same job.

jcape (changes requested):

untrusted/build.rs
nick-mobilecoin: In that case should the feature be `sim` and the default be `[]`?
For getting one started in most environments `sim` seems _nice_, but the main intent is to run on hardware SGX.
Defualting to hardware in a cargo configuration is less likely to trip up multiple users and would only affect initial integrators.  Still might be a bit frustrating for integrators...

untrusted/test_enclave/build.rs
nick-mobilecoin: My initial thought was that this untrusted crate(s) would eventually be utilized to help test the trusted crate(s) behavior as well as any trusted code. 
I was in the mindset that actually building and signing of an enclave would reside in the trusted crate(s) and so to prevent a circular dependency the enclave building logic was crudely repeated here for testing use.

Perhaps,  there should be 3 groups of crates:
* untrusted
* trusted
* sgx_types and enclave building logic

nick-mobilecoin (commented):
accidentally hit the review button instead of the respond, ingore this review...

untrusted/test_enclave/build.rs
jcape: My thinking here is that we're going to end up needing a collection of 4 crates for each `libfoo` provided by the SDK:

1. FFI types (bindgen)
2. FFI methods (bindgen)
3. Rust types (wrapping FFI types)
4. Rust APIs (wrapping FFI methods)

Like, there's zero need to link against an EPID crate in a DCAP world, and we're not doing any switchless stuff in mobilecoin.git (so we shouldn't have those functions in the enclave).

But otherwise, you're right, there is a "last" type for all the build utilities.

untrusted/src/lib.rs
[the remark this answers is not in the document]
nick-mobilecoin: Builder pattern is now  implemented.

> all the Enclave::create() code should go into Enclave::new()
I didn't do this and left the logic in `create()` since it could fail.  Doing a quick scrape on the googles for `new()` it seems like most return the item not a `Result`

untrusted/src/lib.rs
nick-mobilecoin: Looking into this now, found this, https://github.com/intel/linux-sgx/issues/482 
Thinking we'll want to support both files and buffers, not sure how often we've needed to debug in the past, I know I did for the target triple spec file

nick-mobilecoin: using `sgx_create_enclave_from_buffer_ex` is now utilized for the interface

untrusted/src/lib.rs
nick-mobilecoin: Converted to use [sgx_create_enclave_from_buffer_ex](https://download.01.org/intel-sgx/sgx-linux/2.16/docs/Intel_SGX_Developer_Reference_Linux_2.16_Open_Source.pdf#%5B%7B%22num%22%3A232%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C339%2C0%5D)

Note: that the interface had to use a `&mut [u8]`, see https://github.com/mobilecoinfoundation/sgx/blob/3d0c0876bdcf512f5e8db894cd4a624a80d009a0/untrusted/src/lib.rs#L84

untrusted/test_enclave/src/enclave.c
nick-mobilecoin: TBH I was definitely in my C comfort zone when I wrote this.
However digging into this more it runs into the issue the main repo has with the enclave binaries.  Writing the implementation in rust requires one of two approaches:
* Invoking cargo build as a separate step in the build.rs
* Invoking rustc directly and trying to get the correct flags and linking the file into the final lib.

I created https://github.com/mobilecoinfoundation/sgx/issues/4 to capture this.  I think once the trusted FFI gets created and tested we'll have a better idea of what the longer term solution may look like.

untrusted/test_enclave/src/config.xml
nick-mobilecoin: let me know if https://github.com/mobilecoinfoundation/sgx/issues/4 captures this desire, or if you think these changes are taking on unjustifiable debt and should be avoided from the get go.

untrusted/build.rs
nick-mobilecoin: Updated, added a `sw` feature that is part of the default.  The name was chosen to match the value stored in `SGX_MODE`

untrusted/src/lib.rs
```diff
@@ -71,19 +79,32 @@ impl EnclaveBuilder {
...
+            //      let result = unsafe {sgx_create_enclave_from_buffer_ex(...)};
+            //      println!("Post comparing {}", buffer.as_slice() == self.bytes);
+            //
+            // The modification that `sgx_create_enclave_from_buffer_ex()`
+            // makes to the `buffer` is such that if one were to re-use the
+            // modified buffer in another call to
+            // `sgx_create_enclave_from_buffer_ex()` then
+            // `SGX_ERROR_INVALID_ENCLAVE_ID` would be returned.
+            sgx_create_enclave_from_buffer_ex(
+                self.bytes.as_mut_ptr(),
```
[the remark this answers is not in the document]
jcape: If you'd like to maintain the ability to give it a pre-created `Vec<u8>` you can always do `impl From<Vec<u8>> for EnclaveBuilder`, which will give you `EnclaveBuilder::from(existing_vec)[.]...)` as an interface

untrusted/test_enclave/build.rs
nick-mobilecoin: This is now broken up into 3 crates:
* mc-sgx-urts-sys-types
* mc-sgx-urts-sys
* mc-sgx-urts

I went with `sys` for the name to match, from https://doc.rust-lang.org/cargo/reference/build-scripts.html#-sys-packages
For urts I'm not fully sure what types would be independent from the API.

untrusted/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
+resolver = "2"
+members = [
+    "urts",
+    "urts_sys",
+    "urts_sys_types",
+    ]
```
remoun, line 7: Nit: unindent
```suggestion
]
```

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,259 @@
...
+/// See https://doc.rust-lang.org/cargo/reference/environment-variables.html
+fn root_dir() -> PathBuf {
+    PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap())
+}
+
+/// The ld linker to use.  This has to be an ld linker as lld will fail
+/// to link a working enclave
+fn ld_linker() -> String{
+    env::var("LD").unwrap_or_else(|_| String::from("ld"))
+}
```
remoun, line 62: Nit/DRY: Use `into()` instead of `from()`

```suggestion
fn sgx_library_path() -> String {
    env::var("SGX_SDK").unwrap_or_else(|_| DEFAULT_SGX_SDK_PATH.into())
}

/// The value of the environment variable `OUT_DIR`, this must be set.
/// See https://doc.rust-lang.org/cargo/reference/environment-variables.html
fn out_dir() -> PathBuf {
    env::var("OUT_DIR").unwrap().into()
}

/// The root dir of this crate. Will be the value of `CARGO_MANIFEST_DIR`
/// See https://doc.rust-lang.org/cargo/reference/environment-variables.html
fn root_dir() -> PathBuf {
    env::var("CARGO_MANIFEST_DIR").unwrap().into()
}

/// The ld linker to use.  This has to be an ld linker as lld will fail
/// to link a working enclave
fn ld_linker() -> String {
    env::var("LD").unwrap_or_else(|_| "ld".into()))
}
```

untrusted/urts/README.md
```diff
@@ -0,0 +1,49 @@
...
+See https://github.com/intel/linux-sgx#build-the-intelr-sgx-sdk-and-intelr-sgx-psw-package for installation
+instructions. 
+
+The environment variable `SGX_SDK` can be used to specify where the SDK is installed. When unset the location will
+default to `/opt/intel/sgxsdk`
+
+## Features
+
+When the `sw` feature is present the SGX software simulation libraries will be linked in. When the `sw` is absent the
+hardware SGX libraries will be linked in. The `sw` feature is present by default.
```
remoun, line 44: `sim` seems like a more descriptive name for the feature, and matches what we use in the main MC repo

untrusted/urts/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    ///     signed enclave.  This needs to be mutable as the sgx interface will
+    ///     modify the bytes in [EnclaveBuilder::create()].  The `bytes` will
+    ///     not be usable to create another enclave after calling
+    ///     [EnclaveBuilder::create()].  If more than one enclave is needed from
+    ///     the same original `bytes` be sure and copy them to each builder,
+    ///     prior to calling [EncalveBuilder::create()].
+    pub fn new(bytes: &mut [u8]) -> EnclaveBuilder {
+        EnclaveBuilder {
+            bytes,
+            ..Default::default()
```
remoun, line 56: Let's spell out all the fields here
```suggestion
            debug: false,
```

untrusted/src/lib.rs
nick-mobilecoin: This should be addressed now with the separation into 3 crates

untrusted/urts/README.md
nick-mobilecoin: renamed as suggested

untrusted/src/lib.rs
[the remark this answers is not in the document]
nick-mobilecoin: Mad the suggested changes, including the `from()`

untrusted/test_enclave/build.rs
nick-mobilecoin: This is addressed now.  This required renaming the `sim` feature to `hw` and adding similar logic to the test_enclave so one can do `cargo test --features "hw" --features "test_enclave/hw"` to run the tests against the HW SDK

untrusted/test_enclave/README.md
```diff
@@ -0,0 +1,79 @@
...
+2. `enclave.c` and `enclave_t.c` are built into a static library `enclave.a`.
+3. `enclave.a` is linked into a shared library `enclave.so`. This links in the Intel SGX SDK libraries.
+4. `enclave.so` is signed with the `sgx_sign` utility generating `enclave.signed.so`. This is the final binary used for
+   creating an enclave with `sgx_create_enclave()`.
+
+```mermaid
+graph LR
+    A(enclave.edl) -->|edger8r| B(enclave_t.c)
+    B --> D(enclave.a)
+    C(encalve.c) --> D
```
remoun, line 57:
```suggestion
    C(enclave.c) --> D
```

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,271 @@
...
+/// variable `SGX_SDK`.  If this isn't set it will default to
+/// `/opt/intel/sgxsdk`.
+fn sgx_library_path() -> String {
+    env::var("SGX_SDK").unwrap_or_else(|_| DEFAULT_SGX_SDK_PATH.into())
+}
+
+/// The value of the environment variable `OUT_DIR`, this must be set.
+/// See https://doc.rust-lang.org/cargo/reference/environment-variables.html
+fn out_dir() -> PathBuf {
+    env::var("OUT_DIR").unwrap().into()
```
remoun, line 49: Let's use `expect()` for better error messages
```suggestion
    env::var("OUT_DIR").expect("Missing env.OUT_DIR").into()
```

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+/// The root dir of this crate. Will be the value of `CARGO_MANIFEST_DIR`
+/// See https://doc.rust-lang.org/cargo/reference/environment-variables.html
+fn root_dir() -> PathBuf {
+    env::var("CARGO_MANIFEST_DIR").unwrap().into()
```
remoun, line 55:
```suggestion
    env::var("CARGO_MANIFEST_DIR").expect("Missing env.CARGO_MANIFEST_DIR").into()
```

untrusted/test_enclave/build.rs
```diff
@@ -0,0 +1,271 @@
...
+    env::var("LD").unwrap_or_else(|_| "ld".into())
+}
+
+fn sgx_library_suffix() -> &'static str {
+    match () {
+        #[cfg(feature = "hw")]
+        () => "",
+        #[cfg(not(feature = "hw"))]
+        () => "_sim",
+    }
```
remoun, line 70: The `match` seems unnecessary here
```suggestion
    #[cfg(feature = "hw")]
    ""
    #[cfg(not(feature = "hw"))]
    "_sim"
```

Or better yet, make it a constant (with 2 `cfg`-gated declarations)

untrusted/urts/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[dependencies]
+mc-sgx-urts-sys = { path = "../urts_sys", default-features = false }
+
+[features]
+hw = ["mc-sgx-urts-sys/hw"]
+default = []
+
+[dev-dependencies]
+test_enclave = { path = "../test_enclave" }
```
remoun, line 14: Nit: Add newline
```suggestion
test_enclave = { path = "../test_enclave" }

```

untrusted/urts/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    #[test]
+    fn creating_enclave_succeeds() {
+        let builder = EnclaveBuilder::new(ENCLAVE);
+        assert!(builder.create().is_ok());
+    }
+
+    #[test]
+    fn calling_into_a_an_enclave_function_provides_valid_results() {
```
remoun, line 164:
```suggestion
    fn calling_into_an_enclave_function_provides_valid_results() {
```

untrusted/urts_sys/build.rs
```diff
@@ -0,0 +1,46 @@
...
+        .clang_arg(&format!("-I{}/include", sgx_library_path()))
+        .blocklist_type("*")
+        .parse_callbacks(Box::new(bindgen::CargoCallbacks))
+        // Suppressing warnings from tests, see
+        // https://github.com/rust-lang/rust-bindgen/issues/1651
+        .layout_tests(false)
+        .generate()
+        .expect("Unable to generate bindings");
+
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
```
remoun, line 42:
```suggestion
    let out_path = PathBuf::from(env::var("OUT_DIR").expect("Missing env.OUT_DIR"));
```

untrusted/test_enclave/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+```toml
+[dev-dependencies]
+test_enclave = { path = "some/path/to/this/crate" }
+```
+
+Note: In order to disambiguate the term *build* will be used to describe the process of building different enclave
+    artifacts. The term *create* will be reserved for refering to SGX SDK's use of create ala `sgx_create_enclave()`.
+# Table of Contents
+
```
remoun, line 19:
```suggestion

# Table of Contents
```

untrusted/test_enclave/build.rs
nick-mobilecoin: I haven't had a chance to look up the syntax issue
```diff
index 9f964ad..554a37b 100644
--- a/untrusted/urts_sys/build.rs
+++ b/untrusted/urts_sys/build.rs
@@ -12,12 +12,10 @@ fn sgx_library_path() -> String {
 }
 
 fn sgx_library_suffix() -> &'static str {
-    match () {
         #[cfg(feature = "hw")]
-        () => "",
+        ""
         #[cfg(not(feature = "hw"))]
-        () => "_sim",
-    }
+        "_sim"
 }
 
```

```
➜  untrusted git:(feature/untrusted) ✗ cargo test
   Compiling mc-sgx-urts-sys v0.1.0 (/home/nick/git/sgx/untrusted/urts_sys)
   Compiling test_enclave v0.1.0 (/home/nick/git/sgx/untrusted/test_enclave)
error: expected one of `.`, `;`, `?`, `}`, or an operator, found `#`
  --> urts_sys/build.rs:17:9
   |
16 |         ""
   |           - expected one of `.`, `;`, `?`, `}`, or an operator
17 |         #[cfg(not(feature = "hw"))]
   |         ^ unexpected token

error: could not compile `mc-sgx-urts-sys` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```

Converted to constants

untrusted/urts/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+/// value of the enclave will result in failures to the SGX SDK after the
+/// enclave is dropped.
+#[derive(Debug, PartialEq)]
+pub struct Enclave {
+    // The enclave ID, assigned by the SGX interface
+    id: sgx_enclave_id_t,
+}
+
+/// Build an [Enclave] for use with SGX calls.
+#[derive(Default)]
```
remoun, line 28: I don't think we need a `Default` for this builder.

untrusted/urts/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// Create the enclave
+    ///
+    /// Will talk to the SGX SDK to create the enclave.  Once the enclave has
+    /// been created then calls into the enclave can be made by de-referencing
+    /// the enclave.
+    ///
+    /// See
+    /// <https://download.01.org/intel-sgx/sgx-dcap/1.13/linux/docs/Intel_SGX_Enclave_Common_Loader_API_Reference.pdf>
+    /// for error codes and their meaning.
+    pub fn create(mut self) -> Result<Enclave, Error> {
```
remoun, line 71: Let's tag this `must_use`

```suggestion
    #[must_use]
    pub fn create(mut self) -> Result<Enclave, Error> {
```

untrusted/urts_sys_types/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+//
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
```
remoun, line 5: Optional/Equivalent: Merge these
```suggestion
#![allow(non_camel_case_types, non_snake_case, non_upper_case_globals)]
```

untrusted/urts_sys/src/lib.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+// See https://download.01.org/intel-sgx/sgx-dcap/1.13/linux/docs/Intel_SGX_Enclave_Common_Loader_API_Reference.pdf
+//
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
```
remoun, line 6: Optional/Equivalent: Merge these
```suggestion
#![allow(non_camel_case_types, non_snake_case, non_upper_case_globals)]
```

untrusted/test_enclave/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
```
remoun, line 4: Optional/Equivalent: Merge these
```suggestion
#![allow(non_camel_case_types, non_snake_case, non_upper_case_globals)]
```

remoun (approved):
LGTM

untrusted/urts/src/lib.rs
nick-mobilecoin: Good call!, removed

untrusted/urts/src/lib.rs
nick-mobilecoin: I initially added this but had to remove it.
```
➜  untrusted git:(feature/untrusted) ✗ cargo clippy
   Compiling mc-sgx-urts-sys-types v0.1.0 (/home/nick/git/sgx_2/untrusted/urts_sys_types)
   Compiling mc-sgx-urts-sys v0.1.0 (/home/nick/git/sgx_2/untrusted/urts_sys)
    Checking mc-sgx-urts v0.1.0 (/home/nick/git/sgx_2/untrusted/urts)
warning: this function has an empty `#[must_use]` attribute, but returns a type already marked as `#[must_use]`
  --> urts/src/lib.rs:72:5
   |
72 |     pub fn create(mut self) -> Result<Enclave, Error> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(clippy::double_must_use)]` on by default
   = help: either add some descriptive text or remove the attribute
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#double_must_use
```
`Result` types are marked must use, https://doc.rust-lang.org/std/result/#results-must-be-used.  So marking the method results in a warning about double marking.

untrusted/urts/src/lib.rs
```diff
@@ -0,0 +1,197 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+//! Provides rust wrappers for the SGX untrusted runtime system (uRTS) functionality
+
+use mc_sgx_urts_sys::{
+    sgx_create_enclave_from_buffer_ex, sgx_destroy_enclave, sgx_enclave_id_t, sgx_status_t,
+};
+use std::ops::Deref;
+use std::{os::raw::c_int, ptr};
+
+#[derive(PartialEq, Debug)]
```
remoun, line 10: Nit: Sort
```suggestion
#[derive(Debug, PartialEq)]
```

untrusted/urts/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+        let result = unsafe { ecall_add_2(*enclave, 3, &mut sum) };
+        assert_eq!(result, sgx_status_t::SGX_SUCCESS);
+        assert_eq!(sum, 3 + 2);
+    }
+
+    #[test]
+    fn create_enclave_builder_from_vector() {
+        let vector = ENCLAVE.to_vec();
+        assert!(EnclaveBuilder::from(vector).create().is_ok());
+    }
```
remoun, line 180: Nit: Move instantiation tests closer together.
```suggestion

    #[test]
    fn create_enclave_builder_from_vector() {
        let vector = ENCLAVE.to_vec();
        assert!(EnclaveBuilder::from(vector).create().is_ok());
    }

    #[test]
    fn calling_into_an_enclave_function_provides_valid_results() {
        // Note: the `debug()` was added to ensure proper builder behavior of
        // the `create()` method.  It could go away if another test has need
        // of similar behavior.
        let enclave = EnclaveBuilder::new(ENCLAVE).debug(true).create().unwrap();

        let mut sum: c_int = 3;
        let result = unsafe { ecall_add_2(*enclave, 3, &mut sum) };
        assert_eq!(result, sgx_status_t::SGX_SUCCESS);
        assert_eq!(sum, 3 + 2);
    }
```

remoun (approved):

untrusted/urts/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+        // passed to `sgx_create_enclave()`, instead we focus on the
+        // `as c_int` portion maps correctly to 0 or 1
+        let builder = EnclaveBuilder::new(b"");
+        assert_eq!(builder.debug as c_int, 0);
+    }
+
+    #[test]
+    fn when_debug_flag_is_true_it_is_1() {
+        let builder = EnclaveBuilder::new(b"");
+        let builder = builder.debug(true);
```
remoun, line 194: Optional/Equivalent: Merge these declarations
```suggestion
        let builder = EnclaveBuilder::new(b"").debug(true);
```

remoun (dismissed):
LGTM

untrusted/urts/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+            //      println!("Post comparing {}", buffer.as_slice() == self.bytes);
+            //
+            // The modification that `sgx_create_enclave_from_buffer_ex()`
+            // makes to the `buffer` is such that if one were to re-use the
+            // modified buffer in another call to
+            // `sgx_create_enclave_from_buffer_ex()` then
+            // `SGX_ERROR_INVALID_ENCLAVE_ID` would be returned.
+            sgx_create_enclave_from_buffer_ex(
+                self.bytes.as_mut_ptr(),
+                self.bytes.len().try_into().unwrap(),
```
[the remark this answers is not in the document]
nick-mobilecoin: This may have been my rebasing/restacking on the core types, but this was already addressed.
Mainly want to be sure the correct diff is being reviewed 
The comment shows "outdated"
image

core/sys/types/build.rs
```diff
@@ -24,11 +24,15 @@ fn sgx_library_path() -> String {
 
 fn main() {
     let bindings = Builder::default()
-        .header_contents("core_types.h", "#include <sgx_error.h>")
+        .header_contents(
+            "core_types.h",
+            "#include <sgx_error.h>\n#include <sgx_report.h>",
+        )
```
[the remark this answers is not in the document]
nick-mobilecoin: Are you suggesting removing the `mc-sgx-urts-sys-types` crate?
```rust
        .allowlist_type("sgx_enclave_id_t")
        .allowlist_type("sgx_launch_token_t")
        .allowlist_type("sgx_misc_attribute_t")
```
I'm fine with that if that's the desire.  I thought the `core` types were meant to house those types common to both trusted and untrusted code.

untrusted/quote_verify/src/lib.rs
```diff
@@ -16,8 +16,8 @@ const SIGNATURE_COMPONENT_SIZE: usize = 32;
 // The starting byte of the quote report.  The *QE Report* member of the Quote
 // Signature Data Structure.  Table 4 of
 // https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf
-// TODO as more of the quote structure is brought in, derive the start by summation
-//  of previous type/member sizes.
+// TODO as more of the quote structure is brought in, derive the start by
+// summation  of previous type/member sizes.
 const QUOTING_ENCLAVE_REPORT_START: usize = 0x234;
```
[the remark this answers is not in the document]
nick-mobilecoin: no, 0x234(hex) == 564(decimal).  I'm fine converting to dec if we want, but I'm hoping to clean this up with
>  Extract the quote bytes to a Rust type, need to validate the bytes as they are unpacked.

from #17.

Speaking of do we want to convert task lists to issues for easier linking or take it on a case by case basis?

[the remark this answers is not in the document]
jcape: OK, so where I'm going with this is: we need to actually expose all the types that are contained in a quote as their own types, because we're going to be examining all of them, and we need to be able to use them independently.

core/sys/types/build.rs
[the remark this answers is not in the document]
jcape: For example: functions that live in `libsgx_dcap_ql.so` would get their bindgen prototypes in `mc-sgx-dcap-ql-sys`, and functions that live in `libsgx_epid.so` would get their bindgen prototypes in `mc-sgx-epid-sys`. But if they both use, e.g. `sgx_quote_t`, or `sgx_report_t`, then those would both go into `mc-sgx-core-sys-types`, which `-epid-sys` and `-dcap-ql-sys` both depend on.

That said, what I'm suggesting here is that there's a single `.h` file that this library uses, and that file is the one doing the `#include <sgx_foo.h>`s, so we don't need to edit the build script repeatedly.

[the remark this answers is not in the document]
nick-mobilecoin: > what I'm suggesting here is that there's a single .h file that this library uses, and that file is the one doing the #include <sgx_foo.h>s, so we don't need to edit the build script repeatedly.

So you're saying instead of using [header_contents()](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.header_contents) we create a real header file, which we update as needed, and instead use, [header()](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.header)

untrusted/quote_verify/src/lib.rs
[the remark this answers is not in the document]
nick-mobilecoin: Moved the 
> extract quote bytes

to a  specific issue #25 to ensure it doesn't get glossed over and so it's easier to further break down

core/sys/types/build.rs
```diff
@@ -24,11 +24,15 @@ fn sgx_library_path() -> String {
 
 fn main() {
     let bindings = Builder::default()
-        .header_contents("core_types.h", "#include <sgx_error.h>")
```
awygle, line 27: What's the value of doing `.header_contents("core_types.h", "#include <sgx_error.h>")` vs doing `.header("sgx_error.h")`? Is bindgen unable to find `sgx_error.h`?

awygle: I did see that, and I agree with you in that I don't particularly care how we go about this.

awygle (approved):
LGTM.

PR #8, opened 2022-05-18
Add rust bindings for SGX crypto library

nick-mobilecoin:
Initial rust wrapper to show usage of bindings for SGX crypto library.

This provides the basic bindings to show that the sgx_tcrypto library is linked in and can be executed against.  Only one sgx function was provided with an explicit rust wrapper. 

The tests currently rely on std

Follow on work will bring in rust wrappers for the rest of the crypto functionality.

PR #6, opened 2022-05-16
Add trts rust interface

nick-mobilecoin:
Add FFI bindings for the sgx_trts library.

PR #9, opened 2022-05-20
Initial add of tservice crate

nick-mobilecoin:
Initial add of a rust crate to wrap up the SGX tservice library.

### DIFF
--- a/core/sys/types/Cargo.toml
+++ b/core/sys/types/Cargo.toml
@@ -6,4 +6,3 @@ edition = "2021"
 [build-dependencies]
 bindgen = "0.60.1"
 mc-sgx-core-build = { path = "../../build" }
-once_cell = "1.13.0"
--- a/core/sys/types/Cargo.toml
+++ b/core/sys/types/Cargo.toml
@@ -6,3 +6,4 @@ edition = "2021"
 [build-dependencies]
 bindgen = "0.60.1"
 mc-sgx-core-build = { path = "../../build" }
+once_cell = "1.13.0"
--- a/core/sys/types/build.rs
+++ b/core/sys/types/build.rs
@@ -2,7 +2,6 @@
 //! Builds the FFI type bindings for the common SGX SDK types
 
 use bindgen::{callbacks::ParseCallbacks, Builder};
-use once_cell::sync::Lazy;
 
 // Types that don't have an SGX qualifier.
 //
@@ -17,25 +16,23 @@ use once_cell::sync::Lazy;
 //
 // To keep the noise out of the bindings, we use the underlying type and tell
 // bindgen to map directly to `sgx_<name>` version.
-static ALLOWED_UNDERLYING_TYPES: Lazy<Vec<&str>> = Lazy::new(|| {
-    vec![
-        "_status_t",
-        "_target_info_t",
-        "_attributes_t",
-        "_report_t",
-        "_key_request_t",
-    ]
-});
+const ALLOWED_UNDERLYING_TYPES: &[&str] = &[
+    "_status_t",
+    "_target_info_t",
+    "_attributes_t",
+    "_report_t",
+    "_key_request_t",
+];
 
 #[derive(Debug)]
 struct Callbacks;
 
 impl ParseCallbacks for Callbacks {
     fn item_name(&self, name: &str) -> Option<String> {
-        if ALLOWED_UNDERLYING_TYPES.contains(&name) {
-            Some(format!("sgx{}", name))
-        } else if name.starts_with("_sgx") {
+        if name.starts_with("_sgx") {
             Some(name[1..].to_owned())
+        } else if name.starts_with('_') {
+            Some(format!("sgx{}", name))
         } else {
             None
         }
--- a/core/sys/types/build.rs
+++ b/core/sys/types/build.rs
@@ -2,35 +2,64 @@
 //! Builds the FFI type bindings for the common SGX SDK types
 
 use bindgen::{callbacks::ParseCallbacks, Builder};
+use once_cell::sync::Lazy;
+
+// Types that don't have an SGX qualifier.
+//
+// These types are of the form <name> and later typedefed to an `sgx_<name>`.
+//
+// ```C
+//      typedef struct _foo_name {
+//          int a;
+//          float b;
+//      } sgx_foo_name;
+// ```
+//
+// To keep the noise out of the bindings, we use the underlying type and tell
+// bindgen to map directly to `sgx_<name>` version.
+static ALLOWED_UNDERLYING_TYPES: Lazy<Vec<&str>> = Lazy::new(|| {
+    vec![
+        "_status_t",
+        "_target_info_t",
+        "_attributes_t",
+        "_report_t",
+        "_key_request_t",
+    ]
+});
 
 #[derive(Debug)]
 struct Callbacks;
 
 impl ParseCallbacks for Callbacks {
     fn item_name(&self, name: &str) -> Option<String> {
-        match name {
-            "_status_t" => Some("sgx_status_t".to_owned()),
-            _ => None,
+        if ALLOWED_UNDERLYING_TYPES.contains(&name) {
+            Some(format!("sgx{}", name))
+        } else if name.starts_with("_sgx") {
+            Some(name[1..].to_owned())
+        } else {
+            None
         }
     }
 }
 
 fn main() {
     let sgx_library_path = mc_sgx_core_build::sgx_library_path();
-    let bindings = Builder::default()
+    let mut builder = Builder::default()
         .header_contents(
             "core_types.h",
             "#include <sgx_error.h>\n#include <sgx_report.h>",
         )
         .clang_arg(&format!("-I{}/include", sgx_library_path))
         .newtype_enum("_status_t")
-        .blocklist_function("*")
-        .allowlist_type("_status_t")
-        .allowlist_type("sgx_target_info_t")
         .parse_callbacks(Box::new(Callbacks))
         .ctypes_prefix("core::ffi")
-        .generate()
-        .expect("Unable to generate bindings");
+        .allowlist_type("sgx_key_128bit_t");
+
+    for t in ALLOWED_UNDERLYING_TYPES.iter() {
+        builder = builder.allowlist_type(t)
+    }
+
+    let bindings = builder.generate().expect("Unable to generate bindings");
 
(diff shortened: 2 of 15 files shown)


trusted/service/sys/types/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+#![allow(non_camel_case_types, non_snake_case, non_upper_case_globals)]
+
+// Nesting to work around clippy warnings, see
+// https://github.com/rust-lang/rust-bindgen/issues/1470
+#[allow(clippy::missing_safety_doc)]
+mod bindings {
+    // include!(..);
+    include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+}
+pub use bindings::*;
```
remoun, line 13: IIUC we can fold the `allow(clippy...)` into the top level

```suggestion
#![allow(clippy::missing_safety_doc, non_camel_case_types, non_snake_case, non_upper_case_globals)]

include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
```

trusted/service/sys/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "mc-sgx-service-sys"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+mc-sgx-service-sys-types = { path = "types" }
```
remoun, line 7: I see this pattern repeating, but don't understand it. Why not fold the `*-sys-types` crates into `*-sys`? They both use `bindgen`

trusted/service/sys/types/src/lib.rs
nick-mobilecoin: Sorry, I grabbed the wrong base and was missing my signature so force pushed.
Anyway consolidated and also removed that stray `//include!(..);`

trusted/service/sys/Cargo.toml
nick-mobilecoin: My understanding is that some of the types may get re-used in other crates while the functions/linking won't.

Perhaps @jcape can explain it better

.github/workflows/workspace.yaml
```diff
@@ -10,30 +10,30 @@ env:
...
-        with:
-          toolchain: nightly
-          components: rustfmt
-          override: true
-          profile: minimal
+      # actions-rs/toolchain does not support the newer toml file format
+      # see https://github.com/actions-rs/toolchain/issues/126
+      - name: Setup rust toolchain
+        run: rustup show
+        working-directory: ./trusted
```
remoun, line 21: Consider using [`defaults.run`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrun) to default the `working-directory` for this workflow

trusted/service/sys/Cargo.toml
remoun: How about we default to merging `sys-types` into `sys`, until we need to separate them?

nick-mobilecoin:
Attaching the generated bindings
[types.txt](https://github.com/mobilecoinfoundation/sgx/files/8900660/types.txt)
[sys.txt](https://github.com/mobilecoinfoundation/sgx/files/8900667/sys.txt)

nick-mobilecoin:
* develop
  * **PR #23** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/23" target="_blank">Graphite</a> 
    * **PR #2** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/2" target="_blank">Graphite</a> 
      * **PR #24** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/24" target="_blank">Graphite</a> 
        * **PR #6** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/6" target="_blank">Graphite</a> 
          * **PR #26** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/26" target="_blank">Graphite</a> 
            * **PR #8** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/8" target="_blank">Graphite</a> 
              * **PR #9** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/9" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/9?utm_source=stack-comment).

trusted/service/sys/Cargo.toml
remoun: On further thought, keeping `sys-types` separate SGTM

remoun (approved):

trusted/service/sys/Cargo.toml
nick-mobilecoin: sounds good thanks!

trusted/service/sys/types/src/lib.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+//! FFI types for the SGX SDK trusted service library (tservice).
+
```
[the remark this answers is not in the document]
nick-mobilecoin: Good catch! Added this, also note that I had to update to [use_core](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.use_core) to use `core::marker` over `std::marker`

core/sys/types/Cargo.toml
```diff
@@ -6,3 +6,4 @@ edition = "2021"
 [build-dependencies]
 bindgen = "0.60.1"
 mc-sgx-core-build = { path = "../../build" }
+once_cell = "1.13.0"
```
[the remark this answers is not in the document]
nick-mobilecoin: I played with this a little bit.  `std::sync::Once` needs 2 static variables:
* The actual variable
* An initialization one
The actual variable still has to be staticly initializable which a vector isn't :(.  
I could use array but it requires specifying the size of the array, which seems to be a bit of a maintenence burden, maybe I'm just nitting over one extra value to keep up to date.

core/sys/types/build.rs
```diff
@@ -2,35 +2,65 @@
...
+// bindgen to map directly to `sgx_<name>` version.
+static ALLOWED_UNDERLYING_TYPES: Lazy<Vec<&str>> = Lazy::new(|| {
+    vec![
+        "_status_t",
+        "_target_info_t",
+        "_attributes_t",
+        "_report_t",
+        "_key_request_t",
+    ]
+});
```
remoun, line 28: (may be moot with my other suggestion)

Since this is a fixed immutable list, we can just use a `const` slice
```suggestion
const ALLOWED_UNDERLYING_TYPES: &[&str] = &[
    "_status_t",
    "_target_info_t",
    "_attributes_t",
    "_report_t",
    "_key_request_t",
];
```

core/sys/types/build.rs
```diff
@@ -2,35 +2,65 @@
...
 struct Callbacks;
 
 impl ParseCallbacks for Callbacks {
     fn item_name(&self, name: &str) -> Option<String> {
-        match name {
-            "_status_t" => Some("sgx_status_t".to_owned()),
-            _ => None,
+        if ALLOWED_UNDERLYING_TYPES.contains(&name) {
+            Some(format!("sgx{}", name))
+        } else if name.starts_with("_sgx") {
```
remoun, line 37: When would this branch trigger? I don't see any `_sgx` in the allowlist calls below

core/sys/types/build.rs
```diff
@@ -2,35 +2,65 @@
...
     let sgx_library_path = mc_sgx_core_build::sgx_library_path();
-    let bindings = Builder::default()
+    let mut builder = Builder::default()
         .header_contents(
             "core_types.h",
             "#include <sgx_error.h>\n#include <sgx_report.h>",
         )
         .clang_arg(&format!("-I{}/include", sgx_library_path))
         .newtype_enum("_status_t")
-        .blocklist_function("*")
```
remoun, line 27: This got removed; intended?

core/sys/types/build.rs
```diff
@@ -2,35 +2,65 @@
...
 
 impl ParseCallbacks for Callbacks {
     fn item_name(&self, name: &str) -> Option<String> {
-        match name {
-            "_status_t" => Some("sgx_status_t".to_owned()),
-            _ => None,
+        if ALLOWED_UNDERLYING_TYPES.contains(&name) {
+            Some(format!("sgx{}", name))
+        } else if name.starts_with("_sgx") {
+            Some(name[1..].to_owned())
```
remoun, line 38: Optional/Equivalent: prepend `sgx` if the name starts with `_`
```suggestion
        if name.starts_with("_sgx") {
            Some(name[1..].to_owned())
        } else if name.starts_with("_") {
            Some(format!("sgx{}", name))
```

core/sys/types/Cargo.toml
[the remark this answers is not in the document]
awygle: That's fine, `once_cell` isn't a heavyweight dependency and it's widely used.

awygle (approved):
I have no additional comments over and above remoun's

core/sys/types/build.rs
nick-mobilecoin: Thanks!, I was sure there had to be a nicer way of making a static _list_.

I think I'll leave this declared here for adding to the `allowlist_type()`

core/sys/types/build.rs
nick-mobilecoin: Let me know if you think we should try to document in the `build.rs`.  Since this is core it might justify documenting on top of bindgen docs.
https://rust-lang.github.io/rust-bindgen/allowlisting.html
> If we specify allowlisting rules, then bindgen will only generate bindings to types, functions, and global variables that match the allowlisting rules, or are transitively used by a definition that matches them.

Those `_sgx` types are pulled through transitively

core/sys/types/build.rs
nick-mobilecoin: yes, since we're allowlisting types it's not necessary.
I'm also find adding it back for better explicitness

core/sys/types/build.rs
nick-mobilecoin: used constant slice

core/sys/types/build.rs
nick-mobilecoin: updated similar to suggestion.
clippy complained about `name.starts_with("_")` over `name.starts_with('_')` so used the char version.

PR #10, opened 2022-05-23
Add SHA256 builder

nick-mobilecoin:
Add a SHA256 builder interface to provide piecewise computation of a
SHA256 hash.

### DIFF
--- a/trusted/crypto/src/lib.rs
+++ b/trusted/crypto/src/lib.rs
@@ -117,10 +117,6 @@ impl Sha256Builder {
     pub fn build(self) -> Result<Sha256, Error> {
         let mut hash: sgx_sha256_hash_t = Default::default();
         let result = unsafe { sgx_sha256_get_hash(self.handle, &mut hash) };
-        // The errors from sgx_sha256_close() are success or a NULL pointer.
-        // There isn't much error handling we can do if for some reason
-        // self.handle is a NULL pointer.
-        unsafe { sgx_sha256_close(self.handle) };
         match result {
             sgx_status_t::SGX_SUCCESS => Ok(Sha256 { hash }),
             x => Err(Error::SGX(x)),
@@ -128,20 +124,39 @@ impl Sha256Builder {
     }
 }
 
+impl Drop for Sha256Builder {
+    fn drop(&mut self) {
+        // The errors from sgx_sha256_close() are success or a NULL pointer.
+        // There isn't much error handling we can do if for some reason
+        // self.handle is a NULL pointer.
+        unsafe { sgx_sha256_close(self.handle) };
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
-    use sha2::{self, Digest};
+    use sha2::{self, digest::crypto_common::Output, Digest};
+
+    /// Create hash of `bytes` using the [sha2] crate.  This provides a parallel
+    /// implementation of the SHA256 to compare to the SGX wrapper interface.
+    ///
+    /// # Arguments
+    /// - `bytes` the bytes to create the hash for.
+    ///
+    /// # Returns
+    /// The hash of `bytes`.
+    fn hash_with_sha2(bytes: &[u8]) -> Output<sha2::Sha256> {
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(bytes);
+        hasher.finalize()
+    }
 
     #[test]
     fn run_sha256_804() {
         let bytes: [u8; 3] = [8, 0, 4];
         let hash = Sha256::hash(&bytes).unwrap();
-        let expected = {
-            let mut hasher = sha2::Sha256::new();
-            hasher.update(&bytes);
-            hasher.finalize()
-        };
+        let expected = hash_with_sha2(&bytes);
         assert_eq!(hash.as_slice(), &expected[..]);
     }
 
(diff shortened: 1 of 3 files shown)


trusted/crypto/src/lib.rs
```diff
@@ -65,4 +156,34 @@ mod tests {
...
+        let mut builder = Sha256Builder::new().unwrap();
+        for i in bytes {
+            builder = builder.update(&[i]).unwrap();
+        }
+        let hash = builder.build().unwrap();
+        let expected = {
+            let mut hasher = sha2::Sha256::new();
+            hasher.update(&bytes);
+            hasher.finalize()
+        };
```
remoun, line 172: Consider adding a helper (in this mod) for this pattern

trusted/crypto/src/lib.rs
```diff
@@ -65,4 +156,34 @@ mod tests {
...
+        assert_eq!(hash.as_slice(), &expected[..]);
+    }
+
+    #[test]
+    fn build_sha256_1066() {
```
remoun, line 177:
```suggestion
    fn build_sha256_1966() {
```

trusted/crypto/src/lib.rs
```diff
@@ -37,6 +56,78 @@ impl Sha256 {
...
+    ///
+    /// This function will return an error if there is an internal library error
+    /// in SGX.
+    pub fn build(self) -> Result<Sha256, Error> {
+        let mut hash: sgx_sha256_hash_t = Default::default();
+        let result = unsafe { sgx_sha256_get_hash(self.handle, &mut hash) };
+        // The errors from sgx_sha256_close() are success or a NULL pointer.
+        // There isn't much error handling we can do if for some reason
+        // self.handle is a NULL pointer.
+        unsafe { sgx_sha256_close(self.handle) };
```
remoun, line 123: IIUC we should move this Close call into an `impl Drop for Sha256Builder`, or else errors from `update()` will result in unclosed handles.

remoun (changes requested):

trusted/crypto/src/lib.rs
nick-mobilecoin: oh good catch!

trusted/crypto/src/lib.rs
nick-mobilecoin: done

trusted/crypto/src/lib.rs
nick-mobilecoin: fixed bytes

trusted/crypto/src/lib.rs
nick-mobilecoin: added a drop method

trusted/crypto/src/lib.rs
```diff
@@ -37,32 +56,137 @@ impl Sha256 {
     }
 }
 
+/// Builds a [Sha256] hash in a piecewise fashion.  This can be used when all
+/// of the data to hash is not available in contiguous memory.
+pub struct Sha256Builder {
+    handle: sgx_sha_state_handle_t,
+}
+
+impl Sha256Builder {
```
nick-mobilecoin, line 65: Any thoughts on this name?  I went with "builder" because of the builder pattern but wasn't sure if something like "hasher" would make more sense.

cc: @remoun since you already looked over this once.

trusted/crypto/src/lib.rs
```diff
@@ -37,32 +56,137 @@ impl Sha256 {
...
+    ///
+    /// # Returns
+    /// The [Sha256] hash of all the `data`, in order, provided in
+    /// [Sha256Builder::update()]
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if there is an internal library error
+    /// in SGX.
+    pub fn build(self) -> Result<Sha256, Error> {
```
nick-mobilecoin, line 117: thoughts on the name `build`, probably depends on if `Builder` is the right name for the struct.
There are `hash` and `finalize` as possibly names.

cc: @remoun since you looked over this once.

trusted/crypto/src/lib.rs
```diff
@@ -16,17 +19,33 @@ impl From<TryFromIntError> for Error {
...
 impl Sha256 {
-    pub const SIZE: usize = 32;
+    const SIZE: usize = 32;
 
+    /// The hash value as a byte slice
     pub fn as_slice(&self) -> &[u8] {
         &self.hash
     }
 
+    /// Compute the hash of `data`.
```
nick-mobilecoin, line 35: With the addition of the `Builder` would it make sense to rename this `from_bytes` or similar?

cc: @remoun since you already looked over this once

jcape:
I don't think we should use Intel's crypto in our higher-level rust code, TBH.

nick-mobilecoin:
> I don't think we should use Intel's crypto in our higher-level rust code, TBH

> We need the FFI, but we don't need to expose it to the higher-level rust code.

I must have misunderstood. I thought we wanted to wrap all of the FFI's with rust.
It sounds like we just want to expose the FFI's for someone else to call into if needed, but don't need to provide the rust style wrappers for them.
In that case it sounds like I should be ensuring the `build.rs` files are exporting all the necessary symbols and then we rustify on an as needed bases for the things we're really calling into.

jcape:
Yeah, sorry for not being clear, the main thing is "make sure the FFI types and functions can be used from rust," moreso than "make sure there is a rust wrapper for any given `do_foo()`"

PR #11, opened 2022-05-26
Add `Enclave::create_report()`

nick-mobilecoin:
Add the ability for an `Enclave` to create a report.

The enclave reporting is needed for attestation.  In particular it's used to generate a quote and the report can only be created from inside the enclave.
This can be seen at https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/SampleCode/QuoteGenerationSample/App/App.cpp

```mermaid
 sequenceDiagram
    participant A as Untrusted
    participant B as libsgx_dcap_ql.so
    participant C as enclave
    participant D as libsgx_tservice.a
    A->>B: sgx_qe_get_target_info()
    B->>A: target_info
    A->>C: Custom function (target_info)
    C-->>D: sgx_create_report(target_info)
    D-->>C: report (return value)
    C->>A: report (return value)
    A->>B: sgx_qe_get_quote_size()
    B->>A: quote size
    A-->B: sgx_qe_get_quote(report, quote size)
```


untrusted/urts/src/lib.rs
```diff
@@ -19,19 +45,51 @@ pub enum Error {
...
 /// enclave is dropped.
-#[derive(Debug, PartialEq)]
 pub struct Enclave {
     // The enclave ID, assigned by the SGX interface
     id: sgx_enclave_id_t,
+    report_fn: Option<EnclaveReportFn>,
+}
+
+impl Enclave {
+    //TODO should this be called "get_report"?  Named create to match the sgx interface?
```
nick-mobilecoin, line 55: Note this todo, would like to remove as part of this review process

untrusted/urts_sys/src/lib.rs
```diff
@@ -3,9 +3,11 @@
 //
 #![allow(non_camel_case_types, non_snake_case, non_upper_case_globals)]
 
+//TODO should report be propagated here or should callers talk to types directly?
```
nick-mobilecoin, line 6: Note this todo, this could stay for a follow on.
I think now or in an immediate follow on may be time to create a "common" types crate.  
Initially I was thinking it would house the types that are used on both sides of the enclave interface, but I'm thinking it may be better to keep the 2 sides of the enclave separate at the crate level.
However "common" without guiding rules quickly blows up.

untrusted/urts/src/lib.rs
```diff
@@ -193,4 +269,53 @@ mod tests {
...
+                let result =
+                    unsafe { ecall_create_report(**enclave, &mut retval, info, &mut report) };
+                match result {
+                    sgx_status_t::SGX_SUCCESS => match retval {
+                        sgx_status_t::SGX_SUCCESS => Ok(report),
+                        x => Err(Error::SgxStatus(x)),
+                    },
+                    x => Err(Error::SgxStatus(x)),
+                }
+            }))
```
remoun, line 304: This looks like an impl we'd reuse, so how about we promote it to a `pub fn`?

untrusted/urts/src/lib.rs
```diff
@@ -59,6 +118,21 @@ impl EnclaveBuilder {
...
+    ///
+    /// This will be used when calling [Enclave.report()]
+    ///
+    /// # Arguments
+    ///
+    /// * `report_fn` - The function to use for creating [Enclave] reports,
+    ///     [Enclave::report()].  This is `None` by default.  Passing in will
+    ///     prevent creating reports, [Error::NoReportFunction].
+    #[must_use]
+    pub fn report_fn(mut self, report_fn: Option<EnclaveReportFn>) -> EnclaveBuilder {
```
remoun, line 131: When would we not want a `ReportFn`? This might be better as a trait, like [`ReportableEnclave`](https://github.com/mobilecoinfoundation/mobilecoin/blob/5a0be430bece7d52495737746b42e65e5573ee6b/sgx/report-cache/api/src/lib.rs#L59)

untrusted/urts/src/lib.rs
```diff
@@ -19,19 +45,51 @@ pub enum Error {
...
 
 /// Build an [Enclave] for use with SGX calls.
 pub struct EnclaveBuilder {
     // The bytes for the enclave.
     bytes: Vec<u8>,
-
     // `true` if the enclave should be created in debug mode
     debug: bool,
+    // The report function to populate the [Enclave] with.  Defaults to `None`.
+    report_fn: Option<EnclaveReportFn>,
```
remoun, line 92: I'm not convinced this should be optional. It seems better to provide a sensible default, and allow overriding it.

remoun (changes requested):

untrusted/urts/src/lib.rs
nick-mobilecoin: Do you have thoughts on how to provide a default?
I wasn't sure the best place to put it but, https://github.com/mobilecoinfoundation/sgx/blob/feature/enclave_report/untrusted/urts/src/lib.rs#L13
```rust
/// While there is an interface to get an enclave report from inside the
/// enclave, `sgx_create_report()`, there is no standard way to get the
/// report from the untrusted side.
///
/// The [EnclaveReportFn] allows the users of [Enclave]s a general interface
/// to redirect to the specific ecall that the enclave may use for providing
/// back the report.

```
There isn't a common ecall to create the report.  Once in the enclave one can call `sgx_create_enclave()`, but it's up to enclave authors how the report itself is requested through the ecall interface

untrusted/urts/src/lib.rs
nick-mobilecoin: > When would we not want a `ReportFn`?
That's a good question. I think for running production/release enclaves probably almost never.
Getting an enclave up and going even for debug purposes has quite a few boxes to check so I made this optional to lower the barrier to entry a little bit.  I'll admit it may be complicating and over engineering. 

As for the trait, I think it sounds like a good idea, and a good reminder I need to look at them more, but I'm not quite seeing how to do it.  

Defining the trait in urts/src/lib.rs prevents one from implementing it on [Enclave].
Defining the trait in the in progress DCAP quote generation crate requires that crate to know what the "enclave.so" function is.  

The main issue is that the "enclave.so" specific function has to be called and it can be specific to the enclave which means one has to provide the enclave specific function somehow to these more generic interfaces.

untrusted/test_enclave/src/enclave.c
```diff
@@ -1,7 +1,25 @@
...
+/*
+ * Create the report for this enclave instance
+ *
+ * \param target_info The target info to use for creating the report
+ * \param report the report to be filled out.  This is only valid if
+ *      this returns `SGX_SUCCESS`.
+ * \returns The result of `sgx_create_report()`
+ */
+sgx_status_t ecall_create_report(const sgx_target_info_t* target_info, sgx_report_t* report){
+    sgx_report_data_t report_data = { 0 };
```
eranrund, line 23: @jcape correct me if I'm wrong, but I think we want to be able to pass report data?

untrusted/urts/src/lib.rs
nick-mobilecoin: Re-worked this in https://github.com/mobilecoinfoundation/sgx/pull/11/commits/3af723a26b569caa64e7873071c37c4cc7aee359

untrusted/urts/src/lib.rs
```diff
@@ -193,4 +276,38 @@ mod tests {
...
+        let enclave = EnclaveBuilder::new(ENCLAVE)
+            .report_fn(Some(|enclave, target_info| {
+                Ok(test_enclave::enclave_report(**enclave, target_info)?)
+            }))
+            .create()
+            .unwrap();
+
+        let target_info = MaybeUninit::zeroed();
+        let target_info = unsafe { target_info.assume_init() };
+        let report = enclave.create_report(Some(&target_info)).unwrap();
```
remoun, line 302: Why are we passing a zero-value `target_info` instead of `None`?
```suggestion
        let report = enclave.create_report(None).unwrap();
```
... or perhaps another test that passes None

untrusted/urts/src/lib.rs
remoun: I see. Leaving it optional SGTM then.

untrusted/test_enclave/src/lib.rs
```diff
@@ -4,5 +4,41 @@
 /// The test enclave as bytes.
 pub static ENCLAVE: &'static [u8] = include_bytes!(concat!(env!("OUT_DIR"), "/libenclave.signed.so"));
 
-use mc_sgx_urts_sys_types::{sgx_enclave_id_t, sgx_status_t};
+use std::ptr;
+use std::mem::MaybeUninit;
+
+use mc_sgx_urts_sys_types::{sgx_enclave_id_t, sgx_status_t, sgx_report_t, sgx_target_info_t};
```
remoun, line 10: Nit: Sort deps. I thought we had a lint check for this repo?

```suggestion
use mc_sgx_urts_sys_types::{sgx_enclave_id_t, sgx_report_t, sgx_status_t, sgx_target_info_t};
use std::mem::MaybeUninit;
use std::ptr;

```

untrusted/test_enclave/src/lib.rs
```diff
@@ -4,5 +4,41 @@
...
+///     `target_info` to
+///     [`sgx_create_report()`](https://download.01.org/intel-sgx/sgx-linux/2.8/docs/Intel_SGX_Developer_Reference_Linux_2.8_Open_Source.pdf#%5B%7B%22num%22%3A281%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C179.25%2C0%5D).
+pub fn enclave_report(eid: sgx_enclave_id_t, target_info: Option<&sgx_target_info_t>) -> Result<sgx_report_t, sgx_status_t> {
+    let report = MaybeUninit::zeroed();
+    let mut report = unsafe { report.assume_init() };
+    let mut retval: sgx_status_t = sgx_status_t::SGX_SUCCESS;
+    let info = match target_info {
+        Some(info) => info,
+        None => ptr::null(),
+    };
```
remoun, line 35: Optional/Equivalent: Use `unwrap_or_else`

```suggestion
    let info = target_info.unwrap_or_else(|| ptr::null());
```

untrusted/urts/src/lib.rs
remoun: I was thinking we'd have a `ReportableEnclave` trait, but then we'd need a `ReportableEnclaveBuilder`, too. Since we're not currently defining such traits, leaving it as an `Option<EnclaveReportFn>` SGTM

untrusted/test_enclave/src/lib.rs
```diff
@@ -4,5 +4,41 @@
...
+    let result = unsafe { ecall_create_report(eid, &mut retval, info, &mut report) };
+    match result {
+        sgx_status_t::SGX_SUCCESS => match retval {
+            sgx_status_t::SGX_SUCCESS => Ok(report),
+            x => Err(x),
+        },
+        x => Err(x),
+    }
+}
```
remoun, line 44: Nit: Add newline
```suggestion
}

```

untrusted/urts/src/lib.rs
nick-mobilecoin: Added a test case for `None`.
I'll try to rethink the code there, it feels like I'm just exercising the code and not testing it per se. 
Perhaps longer term the test `report_fn` should just provide a means to say what `enclave` and `target_info` were passed to it.

untrusted/test_enclave/src/lib.rs
nick-mobilecoin: Note I undid this change the unwrap returns a `target_info` which means the `or_else` part needs to as well.
I'm guessing I could make some kind of `Option<ptr>` and convert target_info into that and then do an `unwrap_or_else()` there, but unless there is a type already laying around I think the match is probably good enough..??

PR #12, opened 2022-06-02
Add logic to generate an enclave quote

nick-mobilecoin:
Adds an initial attempt at creating an enclave quote.

This creates an enclave quote utilizing the in process quote generation, https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf

This will only work on HW SGX after building and installing the packages,  https://github.com/intel/SGXDataCenterAttestationPrimitives/tree/master/QuoteGeneration#build-the-intelr-sgx-dcap-quote-generation-library-and-the-intelr-sgx-default-quote-provider-library-package.

### DIFF
--- a/untrusted/dcap/src/lib.rs
+++ b/untrusted/dcap/src/lib.rs
@@ -99,7 +99,7 @@ impl Quote {
         if result != quote3_error_t::SGX_QL_SUCCESS {
             // There isn't any corrective action we can take if there is a
             // failure to unload the quoting enclaves
-            print!("Error in cleaning up enclave policy");
+            println!("Error in cleaning up enclaves: {:?}", result);
         }
     }
 
--- a/untrusted/dcap/src/lib.rs
+++ b/untrusted/dcap/src/lib.rs
@@ -51,7 +51,7 @@ impl Quote {
     /// # Arguments
     /// - `enclave` The enclave to generate the quote for.
     pub fn new(enclave: &Enclave) -> Result<Self, Error> {
-        Self::load_in_process_enclaves()?;
+        Self::set_in_process_enclave_paths()?;
         // TODO need to have a common type instead of transmuting these
         let target_info = Self::get_target_info()?;
         let target_info = unsafe { mem::transmute(target_info) };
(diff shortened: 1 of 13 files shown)


untrusted/dcap/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+        let report = enclave.create_report(Some(&target_info))?;
+        let report = unsafe { mem::transmute(report) };
+        let quote = Self::get_quote(report);
+        Self::cleanup_in_process_enclaves();
+        quote
+    }
+
+    fn load_in_process_enclaves() -> Result<(), Error> {
+        //TODO this should be guarded by a feature and this should only be done
+        //  once, maybe lazy_static
```
remoun, line 67: https://docs.rs/once_cell seems like a better fit for that

nick-mobilecoin: nice call out, I like how you can have local variables with it

untrusted/dcap/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+use mc_sgx_urts::Enclave;
+use std::ffi::CString;
+use std::mem;
+use std::mem::MaybeUninit;
+
+/// Errors generating quotes
+#[derive(Debug, PartialEq)]
+pub enum Error {
+    // An error provided from the SGX SDK
+    SgxStatus(quote3_error_t),
```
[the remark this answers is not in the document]
nick-mobilecoin: Looks like there was an issue in bindgen, https://github.com/rust-lang/rust-bindgen/pull/2144, which was released yesterday... 
Brought the update to bindgen in to get hex output for now.
Created #13, I think it will most likely take some libclang parsing of the enums ourselves to generate the lookup table so leaving it for a follow on issue.

nick-mobilecoin:
Converted this to a draft to focus on #14 for minimal FFI support

NotGyro (approved):
Looks good to me. The one thing I might mention is that the "TODO this should be guarded by a feature" might be easy to knock out quickly, but if that's outside of the scope of this pull request, that's understandable.

PR #21, opened 2022-07-07
Add ability to verify enclave report body

nick-mobilecoin:
The enclave report body can now be verified using the attestation key in
the quote structure.


nick-mobilecoin:
* develop
  * **PR #21** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/21" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/21?utm_source=stack-comment).

untrusted/quote_verify/src/lib.rs
```diff
@@ -52,15 +88,53 @@ impl Quote {
...
+    /// Gets the quote header and enclave report body.
+    fn get_header_and_enclave_report_body(&self) -> &[u8] {
+        &self.bytes[..QUOTE_HEADER_SIZE + ENCLAVE_REPORT_SIZE]
+    }
+
+    /// Get the public signing key for the enclave report body (and header)
+    fn get_pub_key(&self) -> Result<Pk, mbedtls::Error> {
+        let mut start = ATTESTATION_KEY_START;
+
+        // `from_binary()` is a wrapper around `mbedtls_mpi_read_binary()`, it
```
nick-mobilecoin, line 115: I'm up for suggestions with regard to these comments.  They may be to verbose.
Initial development I had `unwrap()` and then when I went and converted to `?` I tried to see if I could test the logic and thus the comments.

untrusted/quote_verify/src/lib.rs
```diff
@@ -87,8 +161,8 @@ impl Quote {
     // TODO strict DER requires an extra 0 prefix on the r value when the r
     //  value's most significant bit is set.  This is because `r` should be
     //  unsigned, but the underlying ASN.1 of DER uses an integer [signed].
-    fn get_der_signature(&self) -> Vec<u8> {
-        let mut start = QUOTING_ENCLAVE_SIGNATURE_START;
+    fn get_der_signature(&self, offset: usize) -> Vec<u8> {
```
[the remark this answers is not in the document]
remoun: This seems fine for an internal method

untrusted/quote_verify/src/lib.rs
remoun: I don't see much value in these comments; when would a reader care about the likelihood of failure?

remoun (approved):
LGTM, modulo the extra comments

untrusted/quote_verify/src/lib.rs
```diff
@@ -87,8 +145,8 @@ impl Quote {
     // TODO strict DER requires an extra 0 prefix on the r value when the r
```
awygle, line 145: Is this OK to leave as a TODO? What are the consequences of hitting this case in this context? Seems like we'll fail to verify the signature and thus fail the report, which seems like a significant issue to me.

awygle:
I reviewed this because my assigned PR depended on it. LGTM except for the one note I left.

untrusted/quote_verify/src/lib.rs
awygle: I think I'd at least like to see an issue for it. The problem with TODOs is they don't get tracked against, like, releases, so we can easily end up shipping without addressing them. My first-order reading of this TODO is "if we ship with this it will be Bad because things will apparently-randomly fail to validate", hence my concern.

nick-mobilecoin: This specific instance should be addressed by #27.  I'm fine moving forward with TODO's needing an issue (or perhaps PR.? not sure on the PR part as they could get changed before merge, but so could an issues text.

PR #22, opened 2022-07-08
Verify the attestation key

nick-mobilecoin:
Verifies the attestation key by comparing the SHA256 of it and the QE
authentication data with the QE report data


nick-mobilecoin:
* develop
  * **PR #21** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/21" target="_blank">Graphite</a> 
    * **PR #22** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/22" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/22?utm_source=stack-comment).

untrusted/quote_verify/src/lib.rs
```diff
@@ -185,14 +244,19 @@ impl Quote {
 pub enum Error {
```
remoun, line 244: Optional: Add `derive(displaydoc::Display)`

untrusted/quote_verify/src/lib.rs
```diff
@@ -185,14 +244,19 @@ impl Quote {
...
+
     /// Failure to verify a Signature
     Signature(mbedtls::Error),
+
     /// A failure to convert a binary key into an mbedtls version.
     /// This is unlikely to happen as it should only happen if there is an
     /// error in the FFI or if there is a failure to malloc.
     Key(mbedtls::Error),
+
+    AttestationKey,
```
remoun, line 259: Let's add rustdoc for this
```suggestion
    /// Invalid attestation key in quote.
    AttestationKey,
```

remoun (approved):
LGTM

untrusted/quote_verify/src/lib.rs
nick-mobilecoin: Added `Display`.  I had to convert a multi-line comment into a block comment, not sure if there are any thoughts on mixing these in the same type.

> Multi-line comments are disabled by default by displaydoc. Please consider using block doc comments (/** */)

The error message is a bit misleading in that multi-line comments aren't supported, the subsequent lines are ignored, https://github.com/yaahc/displaydoc#details

untrusted/quote_verify/src/lib.rs
nick-mobilecoin: done

untrusted/quote_verify/src/lib.rs
```diff
@@ -61,6 +63,28 @@ const QUOTING_ENCLAVE_REPORT_START: usize = ATTESTATION_KEY_START + KEY_SIZE;
 // https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf
 const QUOTING_ENCLAVE_SIGNATURE_START: usize = QUOTING_ENCLAVE_REPORT_START + ENCLAVE_REPORT_SIZE;
 
+// The starting byte of the signature for the quote. *QE Report Signature* of
```
awygle, line 66: This comment is incorrect (copied from above constant).

untrusted/quote_verify/src/lib.rs
```diff
@@ -75,6 +99,8 @@ pub struct Quote {
     bytes: Vec<u8>,
 }
 
+impl Quote {}
```
awygle, line 102: I think this is extraneous?

untrusted/quote_verify/src/lib.rs
```diff
@@ -103,6 +129,40 @@ impl Quote {
...
+        let key = &self.bytes[ATTESTATION_KEY_START..ATTESTATION_KEY_START + KEY_SIZE];
+        hasher.update(key);
+
+        let authentication_data = self.get_qe_authentication_data();
+        hasher.update(authentication_data);
+
+        let hash = hasher.finalize();
+        let report_data = &self.bytes
+            [QUOTING_ENCLAVE_REPORT_DATA_START..QUOTING_ENCLAVE_REPORT_DATA_START + hash.len()];
+
```
awygle, line 145: Is there any reason to also verify that the last 32 bytes of the Report Data are 0x00's? It shouldn't be relevant to the verification of the attestation key but it is described in Table 4.

awygle (changes requested):

untrusted/quote_verify/src/lib.rs
nick-mobilecoin: Add logic and tests for this.

I'm a bit torn if this is perhaps making the check too strict in the event Intel wants to use those 32 bytes for something.  Since we're comparing a SHA256 digest, it seems unlikely we would get a valid digest and unexpected values in there.
But I didn't feel strongly enough to push back on it.

awygle (approved):
Thanks!

PR #23, opened 2022-07-11
Add FFI for `sgx_status_t`

nick-mobilecoin:
Generates the FFI bindings for the `sgx_status_t`. This binding uses the
`newtype_enum` to ensure an enum per possible value while preventing
undefined behavior if an unknown return type is provided.


nick-mobilecoin:
attaching bindings for the `sgx_status_t` value
[bindings.txt](https://github.com/mobilecoinfoundation/sgx/files/9085259/bindings.txt)

jcape (approved):
We can dump the `mc-sgx-core-sys` crate, otherwise looks good---we'll want to add additional types to this crate, but this is a good start.

nick-mobilecoin:
> We can dump the `mc-sgx-core-sys` crate, otherwise looks good---we'll want to add additional types to this crate, but this is a good start.

I'll drop the `mc-sgx-core-sys` crate, think i was just copying the other layouts, but it is just noise here.
As for the additional types my thought was to add them on an as needed bases, similar to https://github.com/mobilecoinfoundation/sgx/pull/2/files#diff-8c75269a3c404dce7baf3463401473b295282e8fb022b384144bb350e256b3a1

jcape:
Also, you need to set the C-types location, it's putting `std::os::raw::c_uint` into `sgx_status_t`

remoun (approved):

PR #28, opened 2022-07-19
Verify the quote certificate chain

nick-mobilecoin:


untrusted/quote_verify/src/lib.rs
```diff
@@ -103,6 +117,28 @@ impl Quote {
...
+        // verifying from the root.
+        pems.reverse();
+
+        let mut key_bytes = root_cert.public_key().raw.to_vec();
+        for pem in pems.iter() {
+            let cert = pem.parse_x509()?;
+            let (_, key) = SubjectPublicKeyInfo::from_der(key_bytes.as_slice())?;
+            cert.verify_signature(Some(&key))?;
+            key_bytes = cert.public_key().raw.to_vec();
+        }
```
nick-mobilecoin, line 138: Not liking this logic here.  I was fighting getting to the certificates and persisting them around correctly.  Since the logic is:
```mermaid
graph LR
    A(buffer) --> B(PEM) --> C(X509)
```
The lifetime of the buffer doesn't carry through to the X 509 cert structure.

awygle (approved):
This looks good! It might be worth citing some source for the certificate ordering, if we have one, or mentioning it's empirically determined, but I wouldn't block on that.

NotGyro:
Tests are failing in CI, but it looks like that's just because somebody needs to build this locally so that `Cargo.lock` gets updated, and then commit the updated Cargo.lock. The specific error message is `error: the lock file /__w/sgx/sgx/untrusted/Cargo.lock needs to be updated but --locked was passed to prevent this`

I could try this, but I don't know if I'd be stepping on anything important by committing to this branch right now.

nick-mobilecoin:
> I could try this, but I don't know if I'd be stepping on anything important by committing to this branch right now.

Should be good now, if I did it all correctly...

NotGyro (approved):
Looks good! I like the concise but informative comments.

PR #38, opened 2022-08-02
SGX Capable FFI

jcape:
- WIP
- Add capable-ffi-types
- Finish off capable/ffi/types
- Add start of mc-sgx-capable-ffi-sys crate.
- Build Cleanup
- Update README for ffi-sys crate.


capable/ffi/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+description = '''
+This crate exposes FFI bindings for the SGX SDK's `libsgx_capable` libraries.
+'''
+
+[dependencies]
+mc-sgx-capable-ffi-types = { path = "types" }
+mc-sgx-capable-ffi-sys = { path = "sys" }
+
+[build-dependencies]
+bindgen = ""
```
nick-mobilecoin, line 19: Should this be specifying a version?
```suggestion
bindgen = "0.60.1"
```

capable/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "mc-sgx-capable"
+version = "0.1.0"
+edition = "2021"
+authors = ["MobileCoin"]
```
nick-mobilecoin, line 5: Do we want the other values populated here
```suggestion
authors = ["MobileCoin"]
rust-version = "1.62.1"
license = "Apache-2.0"
readme = "README.md"
repository = "https://github.com/mobilecoinfoundation/sgx"
```

capable/ffi/sys/build.rs
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+
+//! Builds the FFI type bindings for the types used by libsgx_capable.{so,a}.
```
nick-mobilecoin, line 3: I think we said the `so` wasn't going to be used
```suggestion
//! Builds the FFI type bindings for the types used by libsgx_capable.a.
```

capable/ffi/sys/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2022 MobileCoin Inc.
+
+//! FFI bindings for methods in `libsgx_capable.so`.
+
+#![no_std]
```
nick-mobilecoin, line 5: if this is only from the untrusted side does this need to be no_std?
```suggestion
```

capable/ffi/types/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2022 MobileCoin Inc.
+
+//! FFI type bindings for the types used by libsgx_capable.{so,a}.
+
+#![no_std]
```
nick-mobilecoin, line 5: Similar to the sys crate, not sure no_std is needed
```suggestion
```

capable/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        let result = 2 + 2;
+        assert_eq!(result, 4);
+    }
+}
```
nick-mobilecoin, line 8: Should this be empty, don't see much harm leaving a test in here but it might be noise in test output
```suggestion
```

nick-mobilecoin (approved):
There are a couple of empty `README.md` files that should probably either go away or have contents added to them
* `capable/README.md`
* `capable/ffi/README.md`

capable/Cargo.toml
jcape: Yeah, I basically ignored this because this crate isn't currently built out yet.

capable/src/lib.rs
jcape: I'm fine to leave this here for now, the crate isn't actually hooked into anything yet.

capable/ffi/types/src/lib.rs
awygle: Alternately, if we don't need `std`, why not be `no_std`? What does it cost us?

awygle (approved):
This looks good to me.

capable/ffi/Cargo.toml
jcape: (same as below, this crate isn't actually hooked into anything)

capable/ffi/sys/src/lib.rs
jcape: Thanks, this is a good catch---leaving `std` in will cause enclave builds to break if they're depending on this lib.

capable/ffi/types/src/lib.rs
jcape: `foo-types` crates in this repo should always be `no_std` (I left it here), so we can use those structures inside the enclave. Actual FFIs may or may not, depending on whether the libs they're linking to are "capable" of it... 🥁

PR #41, opened 2022-08-02
Consolidate ParseCallbacks and Builder to common

nick-mobilecoin:
Adds an `SGXParseCallbacks` struct that can be used in a `Builder`.
Adds a function, `sgx_builder()`  to provide a pre-configured `Builder`

### DIFF
--- a/core/build/src/lib.rs
+++ b/core/build/src/lib.rs
@@ -2,9 +2,56 @@
 
 #![doc = include_str!("../README.md")]
 
+use bindgen::{callbacks::ParseCallbacks, Builder, EnumVariation};
 use std::{env, path::PathBuf};
 
 static DEFAULT_SGX_SDK_PATH: &str = "/opt/intel/sgxsdk";
+
+/// Provides a default [bindgen::callbacks::ParserCallbacks::item_name]
+/// implementation that works with most SGX types.
+///
+/// # Arguments
+/// * `name` - The name of the type to determine the bindgen name of.
+pub fn sgx_item_name(name: &str) -> Option<String> {
+    if name.starts_with("_sgx") {
+        Some(name[1..].to_owned())
+    } else if name.starts_with('_') {
+        Some(format!("sgx{}", name))
+    } else {
+        None
+    }
+}
+
+/// Returns a builder configured with the defaults for using bindgen with the
+/// SGX libraries.
+pub fn sgx_builder() -> Builder {
+    Builder::default()
+        .derive_copy(true)
+        .derive_debug(true)
+        .derive_default(true)
+        .derive_eq(true)
+        .derive_hash(true)
+        .derive_ord(true)
+        .derive_partialeq(true)
+        .derive_partialord(true)
+        .default_enum_style(EnumVariation::Consts)
+        .prepend_enum_name(false)
+        .use_core()
+        .ctypes_prefix("core::ffi")
+        .allowlist_recursively(false)
+}
+
+/// SGXParseCallbacks to be used with [bindgen::Builder::parse_callbacks]
+///
+/// This provides a default implementation for most of the SGX libraries
+#[derive(Debug)]
+pub struct SGXParseCallbacks;
+
+impl ParseCallbacks for SGXParseCallbacks {
+    fn item_name(&self, name: &str) -> Option<String> {
+        sgx_item_name(name)
+    }
+}
 
 /// Return the SGX library path.
 ///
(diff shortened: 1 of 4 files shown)


nick-mobilecoin:
* develop
  * **PR #41** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/41" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/41?utm_source=stack-comment).

NotGyro (approved):
This is a good consolidation/cleanup, the resulting code is easier to follow. Looks good to me!

core/build/src/lib.rs
```diff
@@ -2,10 +2,57 @@
...
+        .use_core()
+        .ctypes_prefix("core::ffi")
+        .allowlist_recursively(false)
+}
+
+/// SGXParseCallbacks to be used with [bindgen::Builder::parse_callbacks]
+///
+/// This provides a default implementation for most of the SGX libraries
+#[derive(Debug)]
+pub struct SGXParseCallbacks;
```
samdealy, line 48: The codebase normally uses `Sgx...` rather than `SGX...`. Unless there's a reason I'm missing, could you rename this struct to `SgxParseCallbacks`?

core/build/src/lib.rs
```diff
@@ -2,10 +2,57 @@
...
 use std::{env, path::PathBuf};
 
 static DEFAULT_SGX_SDK_PATH: &str = "/opt/intel/sgxsdk";
 
+/// Provides a default [bindgen::callbacks::ParserCallbacks::item_name]
+/// implementation that works with most SGX types.
+///
+/// # Arguments
+/// * `name` - The name of the type to determine the bindgen name of.
+pub fn sgx_item_name(name: &str) -> Option<String> {
```
samdealy, line 15: Nit: Perhaps rename this function to something more descriptive / uses a verb? I'm not sure what it does by just looking at it. Would `add_sgx_item` be an acceptable name?

core/ffi/types/build.rs
```diff
@@ -1,61 +1,37 @@
...
-        .parse_callbacks(Box::new(Callbacks))
-        .ctypes_prefix("core::ffi")
-        .use_core()
-        .allowlist_type("sgx_key_128bit_t");
-
-    for t in ALLOWED_UNDERLYING_TYPES.iter() {
-        builder = builder.allowlist_type(t)
-    }
+        .parse_callbacks(Box::new(SGXParseCallbacks))
+        .allowlist_type("sgx_key_128bit_t")
```
samdealy, line 16: What's the reasoning behind adding an explicit `allowlist_type` for each type instead of keeping the for loop? The for loop makes it slightly easier to add a new type becasue you don't need to type out `allowlist_type` again, but I guess this way is slightly more readable?

samdealy (changes requested):

core/ffi/types/build.rs
nick-mobilecoin: I'm fine reverting, not really invested one way or the other.
 did this as I chewed on this comment, https://github.com/mobilecoinfoundation/sgx/pull/9#discussion_r927103062

I think my thought was to mimic the other crates that are being more explicit about the types, but perhaps they would be better served having a static slice and we make a common utility to add a slice of allowed types

core/build/src/lib.rs
nick-mobilecoin: Good call out, fixed!

core/build/src/lib.rs
nick-mobilecoin: I went with `sgx_normalize_item_name()`.  I think that better describes what's going on.

nick-mobilecoin: all though the "sgx" prefix may be my C namespace mentality coming in, so feel free to suggest a better name there

core/ffi/types/build.rs
samdealy: SGTM.

samdealy (approved):

core/ffi/types/build.rs
nick-mobilecoin: Thinking about it more, I agree the const slice is nicer, I also sorted them, should have done that earlier sorry, but now I know clion's vim bindings support `sort`

nick-mobilecoin:
**Graphite Merge Job**

Current status: 🔄 Waiting for checks to pass

This pull request is currently waiting for CI and mergeability checks to pass before being merged.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/41?utm_source=stack-comment).


Job Reference: awrIFR4yxeX1nu0a41G3

PR #45, opened 2022-08-08
Make urts functions explicit

nick-mobilecoin:
This also normalizes the urts crate to use the common SGX builder.rs
functionality.

PR #49, opened 2022-08-09
Make trts functions explicit

nick-mobilecoin:
This also normalizes the trts crate to use the common SGX builder.rs
functionality.

### DIFF
--- a/trts/sys/build.rs
+++ b/trts/sys/build.rs
@@ -5,7 +5,6 @@
 use cargo_emit::{rustc_link_lib, rustc_link_search};
 
 const TRTS_FUNCTIONS: &[&str] = &[
-    "sgx_accept_forward",
     "sgx_is_enclave_crashed",
     "sgx_is_outside_enclave",
     "sgx_is_within_enclave",
--- a/trts/sys/build.rs
+++ b/trts/sys/build.rs
@@ -2,27 +2,42 @@
 
 //! Builds the FFI function bindings for trts (trusted runtime system) of the
 //! Intel SGX SDK
-use bindgen::Builder;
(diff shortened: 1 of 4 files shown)


nick-mobilecoin:
* develop
  * **PR #45** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/45" target="_blank">Graphite</a> 
    * **PR #49** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/49" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/49?utm_source=stack-comment).

trts/sys/build.rs
```diff
@@ -2,26 +2,42 @@
 
 //! Builds the FFI function bindings for trts (trusted runtime system) of the
 //! Intel SGX SDK
-use bindgen::Builder;
 use cargo_emit::{rustc_link_lib, rustc_link_search};
 
+const TRTS_FUNCTIONS: &[&str] = &[
+    "sgx_accept_forward",
```
awygle, line 8: Are we sure we want to generate bindings for `sgx_accept_forward`? It's in the `trts.a` but the header which declares it (`trts_inst.h`) is under the `internal/` folder.

awygle (commented):
Question about internal functions being exposed, otherwise looks good.

trts/sys/build.rs
nick-mobilecoin: good catch, removed

awygle (approved):

jcape (approved):
This is fine to go in, but we should think about how we want to handle `sgx_exception_handler_t` in rust..

PR #50, opened 2022-08-09
Make tcrypto functions explicit

nick-mobilecoin:
Also adds READMEs for the tcrypto crates


jcape (approved):

core/build/src/lib.rs
```diff
@@ -44,6 +44,7 @@ pub fn sgx_builder() -> Builder {
         .use_core()
         .ctypes_prefix("core::ffi")
         .allowlist_recursively(false)
+        .parse_callbacks(Box::new(SgxParseCallbacks))
```
awygle, line 47: This showed up in at least one other PR I reviewed today - I assume it'll all shake out as we collapse the PR tree, we just need to make sure it doesn't end up getting merged twice.

tcrypto/sys/types/build.rs
```diff
@@ -2,49 +2,47 @@
...
+    "sgx_ecc_state_handle_t",
+    "sgx_hmac_state_handle_t",
+    "sgx_rsa3072_signature_t",
+    "sgx_rsa_key_type_t",
+    "sgx_rsa_result_t",
+    "sgx_sha1_hash_t",
+    "sgx_sha256_hash_t",
+    "sgx_sha384_hash_t",
+    "sgx_sha_state_handle_t",
+];
```
awygle, line 29: What about types which are declared in `sgx_tcrypto.h` but not used by any of the exported functions? I'm OK with not exporting them but I thought I should bring them up.

* `sgx_hmac_256bit_key_t`
* `sgx_hmac_256bit_tag_t`
* `sgx_generic_ecresult_t`
* `rsa_params_t`

awygle (commented):
Couple of questions about whether we should or should not be binding certain things.

tcrypto/sys/types/build.rs
nick-mobilecoin: I would say types in the header, even if not used should be there.  Longer term @jcape mentioned a desire to automate upgrades, and for this we would most likely be parsing the header.  
There are a few SGX apis that use void pointers, and then later cast to the type so relying on transitive functions probably isn't the best way.

tcrypto/sys/build.rs
```diff
@@ -2,33 +2,85 @@
...
+    "sgx_ecdsa_verify",
+    "sgx_ecdsa_verify_hash",
+    "sgx_free_rsa_key",
+    "sgx_gen_epid_priv_f",
+    "sgx_hmac256_close",
+    "sgx_hmac256_final",
+    "sgx_hmac256_init",
+    "sgx_hmac256_update",
+    "sgx_hmac_sha256_msg",
+    "sgx_init_crypto_lib",
```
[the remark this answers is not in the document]
nick-mobilecoin: removed

tcrypto/sys/build.rs
```diff
@@ -2,33 +2,85 @@
...
+    "sgx_ecc256_close_context",
+    "sgx_ecc256_compute_shared_dhkey",
+    "sgx_ecc256_compute_shared_point",
+    "sgx_ecc256_create_key_pair",
+    "sgx_ecc256_open_context",
+    "sgx_ecdsa_sign",
+    "sgx_ecdsa_verify",
+    "sgx_ecdsa_verify_hash",
+    "sgx_free_rsa_key",
+    "sgx_gen_epid_priv_f",
```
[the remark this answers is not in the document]
nick-mobilecoin: good catch, I was scraping the lib and failed to see this wasn't in the public headers.
removed now

core/build/src/lib.rs
nick-mobilecoin: yes I'm hoping it doesn't collide, I wanted this change and the rename change which happened to be parallel branches

awygle (approved):

PR #53, opened 2022-08-10
Add dcap_ql FFI

nick-mobilecoin:
Adds FFI to the sgx functionality provided in `libsgx_dcap_ql.so`

### DIFF
--- a/core/build/src/lib.rs
+++ b/core/build/src/lib.rs
@@ -35,7 +35,7 @@ pub fn normalize_item_name(name: &str) -> Option<String> {
         .iter()
         .any(|prefix| name.starts_with(prefix))
     {
-        Some(name[1..].to_owned())
+        name.strip_prefix("_").map(str::to_string)
     } else if name.starts_with('_') {
         Some(format!("sgx{}", name))
     } else {
--- a/core/build/src/lib.rs
+++ b/core/build/src/lib.rs
@@ -7,6 +7,19 @@ use std::{env, path::PathBuf};
 
 static DEFAULT_SGX_SDK_PATH: &str = "/opt/intel/sgxsdk";
 
+/// Type name prefixes that need the underscore prefix stripped from them.
+///
+/// For example `_foo_bar` should be `foo_bar`.
+///
+/// While it would be nice to add the `sgx` prefix to any non `_sgx` named type,
+/// the name does not propagate to the types used in the generated bindings.
+/// For example mapping `_foo_bar` to `sgx_foo_bar` would fail because the
+/// following function would still be looking for `foo_bar`.
+/// ```C
+/// void some_function(foo_bar arg);
+/// ```
+const STRIP_UNDERSCORE_PREFIX: &[&str] = &["_sgx", "_tee", "_quote3"];
+
 /// Normalizes a type encountered by bindgen
 ///
 /// Provides a default [bindgen::callbacks::ParserCallbacks::item_name]
@@ -17,8 +30,11 @@ static DEFAULT_SGX_SDK_PATH: &str = "/opt/intel/sgxsdk";
 ///
 /// # Arguments
 /// * `name` - The name of the type to determine the bindgen name of.
-pub fn sgx_normalize_item_name(name: &str) -> Option<String> {
-    if name.starts_with("_sgx") || name.starts_with("_tee") {
+pub fn normalize_item_name(name: &str) -> Option<String> {
+    if STRIP_UNDERSCORE_PREFIX
+        .iter()
+        .any(|prefix| name.starts_with(prefix))
+    {
         Some(name[1..].to_owned())
     } else if name.starts_with('_') {
         Some(format!("sgx{}", name))
@@ -57,7 +73,7 @@ pub struct SgxParseCallbacks;
 
 impl ParseCallbacks for SgxParseCallbacks {
     fn item_name(&self, name: &str) -> Option<String> {
-        sgx_normalize_item_name(name)
+        normalize_item_name(name)
     }
 }
 
--- a/core/build/src/lib.rs
+++ b/core/build/src/lib.rs
@@ -35,7 +35,7 @@ pub fn normalize_item_name(name: &str) -> Option<String> {
         .iter()
         .any(|prefix| name.starts_with(prefix))
     {
-        name.strip_prefix("_").map(str::to_string)
+        name.strip_prefix('_').map(str::to_string)
     } else if name.starts_with('_') {
         Some(format!("sgx{}", name))
     } else {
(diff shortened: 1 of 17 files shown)


nick-mobilecoin:
* develop
  * **PR #53** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/53" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/53?utm_source=stack-comment).

core/build/src/lib.rs
```diff
@@ -17,8 +30,11 @@ static DEFAULT_SGX_SDK_PATH: &str = "/opt/intel/sgxsdk";
...
 /// # Arguments
 /// * `name` - The name of the type to determine the bindgen name of.
-pub fn sgx_normalize_item_name(name: &str) -> Option<String> {
-    if name.starts_with("_sgx") || name.starts_with("_tee") {
+pub fn normalize_item_name(name: &str) -> Option<String> {
+    if STRIP_UNDERSCORE_PREFIX
+        .iter()
+        .any(|prefix| name.starts_with(prefix))
+    {
         Some(name[1..].to_owned())
```
awygle, line 38: Suggest using [strip_prefix](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix) here for clarity of intent (not critical).

jcape (approved):
Looks good to me, we should consider breaking out the DCAP installation action into an "actions" crate.

dcap_ql/sys/types/build.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+//! Builds the FFI type bindings for the dcap quote library of the Intel SGX SDK
+
+use bindgen::callbacks::ParseCallbacks;
+
+const DCAP_QL_TYPES: &[&str] = &[
```
awygle, line 6: This seems to be missing `sgx_target_info_t` (an argument type for `sgx_qe_get_target_info`), `_sgx_ql_pck_cert_id_t` (defined in `sgx_ql_lib_common.h`), `sgx_ql_logging_callback_t` (defined in `sgx_ql_lib_common.h`), and a number of types from `sgx_pce.h`.

It's a bit unclear to me whether `sgx_pce.h` should be included transitively, so that last bit might be irrelevant.

dcap_ql/sys/types/build.rs
```diff
@@ -0,0 +1,56 @@
...
+    "sgx_quote3_error_t",
+    "sgx_ql_path_type_t",
+    "_sgx_ql_request_policy",
+    "_quote3_error_t",
+    "_sgx_ql_qve_collateral_t",
+    "_sgx_ql_qe3_id_t",
+    "_sgx_ql_config_version_t",
+    "_sgx_ql_config_t",
+    "_sgx_ql_qve_collateral_param_t",
```
awygle, line 15: This data type doesn't exist in the version of dcap checked out by the 2.17-reproducible tag on linux-sgx (`dcap_1.11_reproducible`). Is that our relevant DCAP version, or is it 1.14, or the Github master branch?

dcap_ql/sys/types/build.rs
awygle: In light of the below comment I should say that all of these statements are based on the dcap_1.11_reproducible tag checked out by the linux-sgx repo.

awygle (changes requested):
When we answer the question about DCAP version the answer should go into one of the READMEs.

nick-mobilecoin:
> Looks good to me, we should consider breaking out the DCAP installation action into an "actions" crate (so we can use it elsewhere).

I think installing DCAP during CI isn't ideal and we could recover that time by eventually updating the docker image.  

Other than the docker image do we know where it might be used?  
I guess I could see a user using the `sim` feature and running on a vanilla GHA runner.

dcap_ql/sys/types/build.rs
nick-mobilecoin: > This data type doesn't exist in the version of dcap checked out by the 2.17-reproducible tag on linux-sgx 
Are you sure the submodule is up to date? `git submodule update --init --recursive`

Looks like it got added in 1.13.  We are currently targeting 1.14
I'm not sure what the `reproducible` tag is about.  The released version is `2.17.100.3` which shows up on the `2.17` tag, https://github.com/intel/linux-sgx/blob/sgx_2.17/common/inc/internal/se_version.h#L34
The `2.17_reproducible` tag shows `2.17.100.2` https://github.com/intel/linux-sgx/blob/sgx_2.17_reproducible/common/inc/internal/se_version.h#L34

We'll need to figure out where we want to document versions for DCAP and the root SGX

core/build/src/lib.rs
nick-mobilecoin: Done, [afca2be](https://github.com/mobilecoinfoundation/sgx/pull/53/commits/afca2be60616178408231d1fafcd789cfdb633d7)

dcap_ql/sys/types/build.rs
nick-mobilecoin: `sgx_target_info_t` comes from the `mc-sgx-core-sys-types`
I've added the `sgx_pce.h` types since one of them is used directly.

I think we'll need a follow up to update all the readme's for the version of SGX/DCAP being supported.

dcap_ql/sys/types/build.rs
awygle: > We are currently targeting 1.14

Given that version 2.17 of `linux-sgx` (the latest released version) has version 1.11 of dcap as a submodule, I would argue it makes more sense to target that version. Can you fill me in on the reasoning for using 1.14?

nick-mobilecoin: where are you seeing the 2.17 to 1.11 mapping?

https://github.com/intel/linux-sgx/tree/sgx_2.17/external shows [dcap_source @ 68a77a8](https://github.com/intel/SGXDataCenterAttestationPrimitives/tree/68a77a852cd911a44a97733aec870e9bd93a3b86) which has the following commit message:
> Intel(R) SGX DCAP 1.14 Release

Looking at the `1.14` tag in the DCAP repo, https://github.com/intel/SGXDataCenterAttestationPrimitives/tree/DCAP_1.14, it also shows [68a77a8](https://github.com/intel/SGXDataCenterAttestationPrimitives/commit/68a77a852cd911a44a97733aec870e9bd93a3b86)

awygle: Oh that is so weird. My local checkout has 1.11, I even did submodule sync and update. Must be some git thing I messed up.

Never mind, ignore me.

awygle (approved):

PR #57, opened 2022-08-11
Add missing DCAP ql Cargo.toml keys

nick-mobilecoin:
Add missing manifest keys to the Cargo.toml(s) in the DCAP ql crates
to assist with publishing.

Also fixed some comments around clippy exclusions

### DIFF
--- a/dcap_ql/sys/src/lib.rs
+++ b/dcap_ql/sys/src/lib.rs
@@ -1,7 +1,6 @@
 // Copyright (c) 2022 The MobileCoin Foundation
 //! FFI functions for the SGX SDK DCAP ql library.
 
-#![no_std]
 #![allow(non_upper_case_globals, non_camel_case_types, non_snake_case)]
 
 use mc_sgx_core_sys_types::{sgx_report_t, sgx_target_info_t};
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -43,7 +43,7 @@ jobs:
       - run: |
           cargo metadata --no-deps --format-version=1 | \
             jq -r '.packages[].name' | \
-            grep -v -e mc-sgx-core-build -e mc-sgx-dcap-quoteverify-sys -e mc-sgx-quote-verify -e mc-sgx-urts -e mc-sgx-capable-sys | \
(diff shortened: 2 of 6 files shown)


nick-mobilecoin:
* develop
  * **PR #57** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/57" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/57?utm_source=stack-comment).

jcape (approved):

dcap_ql/sys/src/lib.rs
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) 2022 The MobileCoin Foundation
 //! FFI functions for the SGX SDK DCAP ql library.
 
-#![no_std]
```
awygle, line 4: Is there a reason to remove this if the bindings don't require `std`?

awygle (approved):
LGTM. I am not precious about the `#![no_std]` thing, just curious.

dcap_ql/sys/src/lib.rs
nick-mobilecoin: Thinking about this more, I think you're spot on and that there isn't a good reason in this instance.
Doing some digging around it seems that the main downside of using `no_std` in a crate is that it's part of your _contract_ to users.  So if in the future you end up needing to depend on `std` for anything you could potentially break your users.  I don't see that happening for the FFI bindings.  The rust wrappers coming next should probably avoid `no_std` unless they are enclave crates, but for these FFI bindings I think `no_std` would probably be good.

@jcape do you have thoughts here?

I'm merging this as is and we can come back with another PR to standardize since currently I see 4 other sys crates without `no_std`
> ➜  sgx git:(develop) fd lib | xargs rg no_std --files-without-match  
./urts/sys/types/src/lib.rs
./urts/sys/src/lib.rs
./urts/src/lib.rs
./test_enclave/src/lib.rs
./quote_verify/src/lib.rs
./dcap_quoteverify/sys/src/lib.rs
./core/build/src/lib.rs
./capable/sys/src/lib.rs
./capable/src/lib.rs

PR #80, opened 2022-08-16
Add `mc-sgx-core-types` crate

nick-mobilecoin:
Add the initial `mc-sgx-core-types` crate.  It is currently empty and
with the intent to add the desired wrappers in follow on commits

Fixes #78

PR #81, opened 2022-08-16
Add KeyId type

nick-mobilecoin:
Add KeyId type which provides an idiomatic rust wrapper for the
`sgx_key_id_t` type.


nick-mobilecoin:
* develop
  * **PR #80** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/80" target="_blank">Graphite</a> 
    * **PR #81** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/81" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/81?utm_source=stack-comment).

core/sys/types/src/lib.rs
```diff
@@ -11,3 +11,9 @@
 )]
 
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+
+impl Clone for sgx_key_id_t {
+    fn clone(&self) -> sgx_key_id_t {
+        sgx_key_id_t { id: self.id }
+    }
+}
```
jcape, line 19: This should be done by an `mc_sgx_core_build::SgxParseCallbacks::add_derives()` implementation...

jcape (approved):
Looks good generally, but I suggest using `ParseCallbacks::add_derives()` to make everything `#[derive(Clone)]`

core/sys/types/src/lib.rs
nick-mobilecoin: @jcape made an attempt in https://github.com/mobilecoinfoundation/sgx/pull/81/commits/bb63bbdd2766b2b5302bf66824b6fa1111f2ad37 let me know thoughts.
I don't like putting *all* the types in the `build/lib.rs`, perhaps `sgx_builder()` should be updated to take a `non-cloneable` iterator.  I don't like negative args, but it seems to be the narrower scope.

jcape: We should definitely set `derive_copy(false)` in the `sgx_builder()`, that was me simply not thinking.

What we can do is make `SgxParseCallbacks { do_not_clone: Vec<String> }` and `SgxParseCallbacks::new(do_not_clone: impl Iter<Item=&str>)`, and put that into the `Box`.

nick-mobilecoin: next attempt at, https://github.com/mobilecoinfoundation/sgx/pull/81/commits/a824269f393775f66fd3d7760bdbcd5e37b84a19

awygle (approved):
I approve of this as an implementation, if we decide we want a KeyId wrapper.

nick-mobilecoin:
Closing as KeyId doesn't seem like it's needed as a rust type at this time

PR #82, opened 2022-08-16
Add `mc-sgx-dcap-quoteverify-types` crate

nick-mobilecoin:
The `mc-sgx-dcap-quoteverify-types` provides rust wrappers around the
types in `mc-sgx-dcap-quoteverify-sys-types`.

PR #85, opened 2022-08-18
Add rust wrapper for sgx_status_t

nick-mobilecoin:
The rust type `core::types::Error` has been added for encompassing the
`sgx_status_t`.  The previous `core::types::Error` has been renamed to
`core::types::FfiError`

I mostly copied from https://github.com/mobilecoinfoundation/mobilecoin/blob/65791210c87d56867545e549b5328aaeba3df331/sgx/core-types/src/error.rs

Some things to note:
* [failure](https://github.com/rust-lang-deprecated/failure) is deprecated and [thiserror](https://github.com/dtolnay/thiserror) requires `std` so omitted the error messages
* not sure the feeling on losing the u32 repr, this is due to the use of newtypes instead of consts for enums
* Used `From` and mapped to `Unexpected` instead of `TryFrom`

Outstanding question:
* Not sure how this should map function call returns to Errors versus success, but perhaps that was the tryfrom.

### DIFF
--- a/core/types/Cargo.toml
+++ b/core/types/Cargo.toml
@@ -5,3 +5,7 @@ edition = "2021"
 
 [dependencies]
 mc-sgx-core-sys-types = { path = "../sys/types", version = "=0.1.0" }
+serde = { version = "1.0.143", default-features = false, features = ["derive"] }
+
+[dev-dependencies]
+yare = "1.0.1"
--- a/core/types/Cargo.toml
+++ b/core/types/Cargo.toml
@@ -5,7 +5,10 @@ edition = "2021"
 
 [dependencies]
 mc-sgx-core-sys-types = { path = "../sys/types", version = "=0.1.0" }
-serde = { version = "1.0.143", default-features = false, features = ["derive"] }
+serde = { version = "1.0.143", default-features = false, features = ["derive"], optional = true }
+
+[features]
+serde = ["dep:serde"]
 
 [dev-dependencies]
 yare = "1.0.1"
--- a/core/types/src/error.rs
+++ b/core/types/src/error.rs
@@ -1,9 +1,375 @@
 // Copyright (c) 2022 The MobileCoin Foundation
 
+//! SGX Error types
+
+use core::convert::From;
+use mc_sgx_core_sys_types::sgx_status_t;
+use serde::{Deserialize, Serialize};
+
 /// Errors seen when converting to, or from, rust for the SGX types
 #[derive(Debug)]
-pub enum Error {
+pub enum FfiError {
     /// Enum out of range
     /// Happens when a value that is not represented by the known C enum values.
     UnknownEnumValue(i64),
+}
+
+/// A enumeration of SGX errors.
+///
+/// Those listed here are the ones which are identified in the `sgx_status_t`
+/// enum, in order of the actual value. Note that values are grouped
+/// (numerically) into the following general sections:
+///
+///  1. `0x0000-0x0fff`: Generic errors.
+///  2. `0x1000-0x1fff`: Fatal runtime errors.
+///  3. `0x2000-0x2fff`: Enclave creation errors.
+///  4. `0x3000-0x3fff`: Local attestation/report verification errors.
+///  5. `0x4000-0x4fff`: Errors when communicating with the Architectural
+///                      Enclave Service Manager (AESM).  
+///  6. `0x5000-0x5fff`: Errors internal to AESM.  
+///  7. `0x6000-0x6fff`: Errors with the encrypted enclave loader.
+///  8. `0x7000-0x7fff`: Errors with the "SGX Encrypted FS" utility.
+///  9. `0x8000-0x8fff`: Attestation key errors.
+/// 10. `0xf000-0xffff`: Internal (to SGX) errors.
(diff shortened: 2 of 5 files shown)


nick-mobilecoin:
* main
  * **PR #80** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/80" target="_blank">Graphite</a> 
    * **PR #81** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/81" target="_blank">Graphite</a> 
      * **PR #82** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/82" target="_blank">Graphite</a> 
        * **PR #85** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/85" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/85?utm_source=stack-comment).

awygle:
Should we `impl Error` (https://doc.rust-lang.org/std/error/trait.Error.html) for one or both of these error types?

jcape:
> Should we `impl Error` (https://doc.rust-lang.org/std/error/trait.Error.html) for one or both of these error types?

If we do `impl Error`, we'll need to add an `"std"` feature and gate it on that.

core/types/src/error.rs
```diff
@@ -1,9 +1,375 @@
 // Copyright (c) 2022 The MobileCoin Foundation
 
+//! SGX Error types
+
+use core::convert::From;
+use mc_sgx_core_sys_types::sgx_status_t;
+use serde::{Deserialize, Serialize};
```
jcape, line 7: This should be gated on a `"serde"` feature.

core/types/Cargo.toml
```diff
@@ -4,4 +4,8 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-mc-sgx-core-sys-types = { path = "../sys/types", version = "=0.1.0" }
\ No newline at end of file
+mc-sgx-core-sys-types = { path = "../sys/types", version = "=0.1.0" }
+serde = { version = "1.0.143", default-features = false, features = ["derive"] }
```
jcape, line 8: This should be an optional dependency.

core/types/src/error.rs
```diff
@@ -1,9 +1,375 @@
...
+///  4. `0x3000-0x3fff`: Local attestation/report verification errors.
+///  5. `0x4000-0x4fff`: Errors when communicating with the Architectural
+///                      Enclave Service Manager (AESM).  
+///  6. `0x5000-0x5fff`: Errors internal to AESM.  
+///  7. `0x6000-0x6fff`: Errors with the encrypted enclave loader.
+///  8. `0x7000-0x7fff`: Errors with the "SGX Encrypted FS" utility.
+///  9. `0x8000-0x8fff`: Attestation key errors.
+/// 10. `0xf000-0xffff`: Internal (to SGX) errors.
+#[derive(Copy, Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub enum Error {
```
jcape, line 35: `pub enum` should always be `#[non_exhaustive]`, so adding new variants doesn't require a major bump

core/types/src/error.rs
```diff
@@ -1,9 +1,375 @@
...
+///  3. `0x2000-0x2fff`: Enclave creation errors.
+///  4. `0x3000-0x3fff`: Local attestation/report verification errors.
+///  5. `0x4000-0x4fff`: Errors when communicating with the Architectural
+///                      Enclave Service Manager (AESM).  
+///  6. `0x5000-0x5fff`: Errors internal to AESM.  
+///  7. `0x6000-0x6fff`: Errors with the encrypted enclave loader.
+///  8. `0x7000-0x7fff`: Errors with the "SGX Encrypted FS" utility.
+///  9. `0x8000-0x8fff`: Attestation key errors.
+/// 10. `0xf000-0xffff`: Internal (to SGX) errors.
+#[derive(Copy, Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
```
jcape, line 34: I *think* you can do `#[cfg_attr(feature = "serde", derive(Deserialize, Serialize)]` as a second line here and it will do the appropriate feature gating.

jcape (approved):
Looks good, but `serde` should always be optional (e.g. at some point in the future we're going to stop using it in mobilecoin.git), and public enums always need to be `#[non_exhaustive]`.

We should also `#[deny(missing_docs)]` in our idiomatic crates as well (since this is all going on up docs.rs), but that can be a follow-up.

core/types/src/error.rs
nick-mobilecoin: done

core/types/Cargo.toml
nick-mobilecoin: done

core/types/src/error.rs
nick-mobilecoin: done

core/types/src/error.rs
nick-mobilecoin: done

awygle (approved):
LGTM. I am surprised Error isn't in `core`, apparently it's to do with backtraces.

PR #91, opened 2022-08-22
Update the bindgen derive logic

nick-mobilecoin:
Previously many generated types were missing common derive traits like
"Clone", "Debug", etc.  Now most types default to having those traits.
There is a separate list for:
- Types that are enums
- Types that should implement the "Copy" trait.
- Types that are dynamically sized (These implement no traits)

PR #86, opened 2022-08-19
Add rust wrappers for the sgx attributes types

nick-mobilecoin:
Add rust wrappesr for:
- sgx_attributes_t => Attributes
- sgx_misc_select_t => MiscellaneousSelect
- sgx_misc_attribute_t => MiscellaneousAttribute

PR #92, opened 2022-08-23
Add KeyRequest type to wrap sgx_key_request_t

nick-mobilecoin:


nick-mobilecoin:
* main
  * **PR #91** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/91" target="_blank">Graphite</a> 
    * **PR #86** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/86" target="_blank">Graphite</a> 
      * **PR #92** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/92" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/92?utm_source=stack-comment).

core/types/src/key_request.rs
```diff
@@ -0,0 +1,154 @@
...
+impl Default for KeyRequest {
+    fn default() -> Self {
+        KeyRequest(sgx_key_request_t {
+            key_name: 0,
+            key_policy: 0,
+            isv_svn: 0,
+            reserved1: 0,
+            cpu_svn: CpuSvn::default().into(),
+            attribute_mask: Attributes::default().into(),
+            // TODO ideally this should be random
```
nick-mobilecoin, line 33: Not fond of this.  This goes back to our key_id, but to do a _good_ default it seems like we need someway for people to specify an RNG

core/types/src/key_request.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+}
+
+impl KeyRequest {
+    /// Set the key name to use in the key request
+    ///
+    /// # Arguments
+    ///
+    /// * `key_name`: The key name to use
+    pub fn set_key_name(mut self, key_name: u16) -> Self {
```
nick-mobilecoin, line 48: wasn't sure if we wanted to go this far into the types, or if we just wanted more simple newtype wrappers and not worry about how easy it is to construct or manipulate the type at this time.

core/types/src/key_request.rs
jcape: If `Default` is a footgun, then we shouldn't implement it. Based on the setter comment, should we just make a new() method that takes the relevant details?

Like,
```rust
impl KeyRequest {
    pub fn new<R: CryptoRng + RngCore>(csprng: &mut R, key_name: KeyName, ...) -> Self {
        /* ... */
    }
}
```

(I assume that key_policy is supposed to be either an enum or bitflags, even if `key_name` isn't...)

jcape (changes requested):

core/types/src/key_request.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// Set the ISV(Individual Software Vendor) SVN (Security Version Number) of
+    /// the key request
+    ///
+    /// # Arguments
+    ///
+    /// * `isv_svn`: The ISV SVN to use
+    pub fn set_isv_svn(mut self, isv_svn: sgx_isv_svn_t) -> Self {
+        self.0.isv_svn = isv_svn;
+        self
+    }
```
jcape, line 72: Note: we should have real rust types around these SVNs, e.g. `IsvSvn(u16)` to prevent accidentally using one where you intend to use the other.

core/types/src/key_request.rs
nick-mobilecoin: Add `new()` method which takes in a `csprng`.  The new type does not take other parameters, left those for the builder pattern.  I could see an argument for `key_name` and `key_policy` being part of the method.

Also updated `key_name` to be an enum and `key_policy` to support bitwise operations.

core/types/src/key_request.rs
nick-mobilecoin: added real types around the current svns

core/types/src/attributes.rs
```diff
@@ -9,10 +9,47 @@ new_type_wrapper! {
...
+    }
+}
+
 new_type_wrapper! {
     MiscellaneousSelect, sgx_misc_select_t;
 }
 
+// Suppress clippy as the `new_type_wrapper` macro can't derive Default for many
+// of the types.
+#[allow(clippy::derivable_impls)]
```
jcape, line 46: I'd honestly suggest changing this macro to not declare the struct so you can just do the manual derives. I don't know if this is still a thing, but CLion used to have a lot of difficulty figuring out that a type exists if it's defined in a macro...

core/types/Cargo.toml
```diff
@@ -5,10 +5,15 @@ edition = "2021"
 
 [dependencies]
 mc-sgx-core-sys-types = { path = "../sys/types", version = "=0.1.0" }
+rand = { version = "0.8.5", default-features = false }
```
jcape, line 8:
```suggestion
rand_core = { version = "0.6.3", default-features = false }
```

core/types/Cargo.toml
```diff
@@ -5,10 +5,15 @@ edition = "2021"
...
 
 [features]
 serde = ["dep:serde"]
 
 [dev-dependencies]
 yare = "1.0.1"
+
+[dev-dependencies.rand]
+version = "0.8.5"
+features = ["std"]
```
jcape, line 19:
```suggestion
rand = "0.8.5"
```

core/types/src/key_request.rs
```diff
@@ -0,0 +1,273 @@
...
+use crate::{new_type_wrapper, Attributes, MiscellaneousSelect};
+use mc_sgx_core_sys_types::{
+    sgx_config_svn_t, sgx_cpu_svn_t, sgx_isv_svn_t, sgx_key_id_t, sgx_key_request_t,
+    SGX_CPUSVN_SIZE, SGX_KEYID_SIZE, SGX_KEYPOLICY_CONFIGID, SGX_KEYPOLICY_ISVEXTPRODID,
+    SGX_KEYPOLICY_ISVFAMILYID, SGX_KEYPOLICY_MRENCLAVE, SGX_KEYPOLICY_MRSIGNER,
+    SGX_KEYPOLICY_NOISVPRODID, SGX_KEYSELECT_EINITTOKEN, SGX_KEYSELECT_PROVISION,
+    SGX_KEYSELECT_PROVISION_SEAL, SGX_KEYSELECT_REPORT, SGX_KEYSELECT_SEAL,
+    SGX_KEY_REQUEST_RESERVED2_BYTES,
+};
+use rand::{CryptoRng, RngCore};
```
jcape, line 13:
```suggestion
use rand_core::{CryptoRng, RngCore};
```

core/types/src/key_request.rs
```diff
@@ -0,0 +1,273 @@
...
+
+#[repr(u16)]
```
jcape, line 15:
```suggestion
#[non_exhaustive]
#[repr(u16)]
```

core/types/src/key_request.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    /// Derive key with the enclave's CONFIGID
+    pub const CONFIG_ID: KeyPolicy = KeyPolicy(SGX_KEYPOLICY_CONFIGID);
+
+    /// Derive key with the enclave's ISVFAMILYID
+    pub const ISV_FAMILY_ID: KeyPolicy = KeyPolicy(SGX_KEYPOLICY_ISVFAMILYID);
+
+    /// Derive key with the enclave's ISVEXTPRODID
+    pub const ISV_EXTENDED_PROD_ID: KeyPolicy = KeyPolicy(SGX_KEYPOLICY_ISVEXTPRODID);
+}
```
jcape, line 54: Look into the `bitflags` crate for this...

core/types/src/key_request.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+}
+impl ::core::ops::BitAndAssign for KeyPolicy {
+    #[inline]
+    fn bitand_assign(&mut self, rhs: KeyPolicy) {
+        self.0 &= rhs.0;
+    }
+}
+
+new_type_wrapper! {
```
jcape, line 83: I recommend putting all the security version stuff into an `svn.rs` module, since it's going to be used in a lot of places.

core/types/src/key_request.rs
```diff
@@ -0,0 +1,273 @@
...
+    KeyRequest, sgx_key_request_t;
+}
+
+impl KeyRequest {
+    /// Creates a new [KeyRequest].
+    ///
+    /// The [KeyRequest] will be mostly empty and can be built up using the
+    /// setter methods like [KeyRequest::set_isv_svn]
```
jcape, line 127: I didn't think of it before, but the rust way to do this is to have a `KeyRequestBuilder`, e.g.:

`KeyRequestBuilder::new(&csprng).key_name(foo).isv_svn(bar)`

Then you can nuke all the setters and just leave this as an opaque struct (or add getters)

core/types/src/lib.rs
```diff
@@ -3,14 +3,17 @@
 //! Rust wrappers for SGX types
 
 #![no_std]
+#![feature(core_intrinsics)]
```
jcape, line 6: I assume using `bitflags` would eliminate the need for this (separately, if you are using a feature, document why you're using it, so we know when we can remove it---every feature means we're stuck on nightly until it can be removed).

jcape (changes requested):

core/types/src/key_request.rs
```diff
@@ -0,0 +1,273 @@
...
+}
+
+impl KeyRequest {
+    /// Set the key name to use in the key request
+    ///
+    /// # Arguments
+    ///
+    /// * `key_name` - The key name to use
+    pub fn set_key_name(mut self, key_name: KeyName) -> Self {
+        self.0.key_name = ::core::intrinsics::discriminant_value(&key_name);
```
awygle, line 161: If all you want is to turn key_name into a u16, why not just use `key_name as u16`? With `#[repr(u16)]` on the enum, that's safe to do, and doesn't require a language feature.

awygle (changes requested):

core/types/src/attributes.rs
nick-mobilecoin: pulled declaration outside of macro

core/types/Cargo.toml
nick-mobilecoin: done

core/types/Cargo.toml
nick-mobilecoin: done

core/types/src/key_request.rs
nick-mobilecoin: done

core/types/src/key_request.rs
nick-mobilecoin: done

core/types/src/key_request.rs
nick-mobilecoin: brought in `bitflags`

core/types/src/key_request.rs
nick-mobilecoin: created a builder for this.  Note Attributes is still a setter style, I'm thinking since it has a default of all 0s and with only 2 setters it may serve as is.

core/types/src/key_request.rs
nick-mobilecoin: created separate svn module with currently supported svn types

core/types/src/lib.rs
nick-mobilecoin: added a comment.
This is needed for `::core::intrinsics::discriminant_value()`

core/types/src/key_request.rs
nick-mobilecoin: good catch, removed use of `discriminant_value()` in favor of `as u16`

.github/workflows/lint.yaml
```diff
@@ -34,10 +34,13 @@ jobs:
...
         run: cargo install cargo-nono
       # `mc-sgx-quote-verify` should be `no_std`, but it isn't yet.  See #34
+      # `mc-sgx-core-types` should be `no_std`, but cargo nono doesn't parse
+      # correctly, see https://github.com/hobofan/cargo-nono/pull/64 for a
+      # suggested improvement
       - run: |
           cargo metadata --no-deps --format-version=1 | \
             jq -r '.packages[].name' | \
-            grep -v -e mc-sgx-core-build -e mc-sgx-dcap-quoteverify-sys -e mc-sgx-dcap-ql-sys -e mc-sgx-quote-verify -e mc-sgx-urts -e mc-sgx-capable-sys | \
+            grep -v -e mc-sgx-core-types -e mc-sgx-core-build -e mc-sgx-dcap-quoteverify-sys -e mc-sgx-dcap-ql-sys -e mc-sgx-quote-verify -e mc-sgx-urts -e mc-sgx-capable-sys | \
```
jcape, line 43: This should pass with `rand_core`, no?

core/types/src/svn.rs
```diff
@@ -0,0 +1,53 @@
...
+use crate::new_type_accessors_impls;
+use mc_sgx_core_sys_types::{sgx_config_svn_t, sgx_cpu_svn_t, sgx_isv_svn_t, SGX_CPUSVN_SIZE};
+
+#[repr(transparent)]
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Default)]
+pub struct ConfigSvn(sgx_config_svn_t);
+new_type_accessors_impls! {
+    ConfigSvn, sgx_config_svn_t;
+}
+impl ConfigSvn {
```
jcape, line 13:
```suggestion

impl ConfigSvn {
```

jcape (approved):
This is fine as-is, but all these types should have some kind of description of what they're for / what they do that will go into docs.rs.

.github/workflows/lint.yaml
nick-mobilecoin: unfortunatly not
```
➜  sgx git:(feature/key_request) cargo nono check --package mc-sgx-core-types
mc-sgx-core-types: SUCCESS
bitflags: SUCCESS
mc-sgx-core-sys-types: SUCCESS
rand_core: FAILURE
  - Source code contains an explicit `use std::` statement.
   --> /home/nick/.cargo/registry/src/github.com-1ecc6299db9ec823/rand_core-0.6.3/src/error.rs
```

Looking at the file:
```rust
#[cfg(feature = "std")] use std::boxed::Box;   
```
it looks like cargo nono is using some kind of regex parsing or similar, at least for the explicit std check, even manually moving the use to the next line results in the failure.

awygle (approved):
Looks good now. Thanks!

PR #98, opened 2022-08-24
Only build if fast lints pass

jcape:
Makes the "slow" (2-10m) clippy, build, etc. depend on the "fast" rustfmt and markdown-lint jobs completing.

PR #96, opened 2022-08-24
Update GHA workflows:

jcape:

PR #94, opened 2022-08-24
README updates

jcape:
Fixes #88

PR #99, opened 2022-08-24
Add cargo-deny to GHA

jcape:
Add cargo-deny to GHA, will lint if banned packages or copyleft licenses are used.

This also entailed removing `quote_verify` for the time being, since it depended on `ring` and is kind of on hold until the idiomatic rust stuff is done.


jcape:
* main
  * **PR #94** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/94" target="_blank">Graphite</a> 
    * **PR #95** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/95" target="_blank">Graphite</a> 
      * **PR #96** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/96" target="_blank">Graphite</a> 
        * **PR #97** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/97" target="_blank">Graphite</a> 
          * **PR #98** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/98" target="_blank">Graphite</a> 
            * **PR #99** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/99" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/99?utm_source=stack-comment).

deny.toml
```diff
@@ -0,0 +1,49 @@
...
+exceptions = []
+
+[bans]
+multiple-versions = "warn"
+# Lint level for when a crate version requirement is `*`
+wildcards = "deny"
+highlight = "all"
+allow = []
+deny = [
+    { name = "ring" },
```
nick-mobilecoin, line 35: :wrench: should this have a comment saying why it's denied that way if the status changes one will know it's OK to include.  Could also provide an indicator of what other types of crates to avoid.

deny.toml
```diff
@@ -0,0 +1,49 @@
...
+]
+skip = []
+skip-tree = []
```
nick-mobilecoin, line 38: :thought_balloon: Probably not needed with our dependabot reviews, but there is [`allow-build-scripts`](https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html#the-allow-build-scripts-field-optional), which can avoid crates adding a build script that isn't looked at.

nick-mobilecoin (approved):

deny.toml
awygle: Yeah just on a personal level I'm curious about why this is forbidden (although the license situation _does_ look... complicated).

awygle (approved):
This looks fine. I do want to ask one thing, which is - what do you think about keeping the `quote_verify` crate in the repo but not the workspace? We intend to include it eventually, so removing it whole-hog seems like something we might just have to undo again later.

jcape:
> This looks fine. I do want to ask one thing, which is - what do you think about keeping the `quote_verify` crate in the repo but not the workspace? We intend to include it eventually, so removing it whole-hog seems like something we might just have to undo again later.

I'm OK with having to pull it back later, it lives forever in git's history. When we go to bring it back, we'll have a better understanding of what we need out of it, and it won't require maintenance in the interim...

deny.toml
jcape: Hmm, interesting. I'm a bit torn. I think it's an abstract good that we have documented which dependencies have build scripts (so we can allowlist them), but as you say, there may not be a practical benefit to doing so...

I think we should leave it out for now, based on thorough dependabot reviews.

deny.toml
nick-mobilecoin: Thanks for the issue link, should we link to that issue in the toml file?  I think it will be hard to clearly capture in one or two sentences the disparity in opinions on that issue.

jcape: Yep, done

PR #112, opened 2022-08-29
Create a `Report` rust wrapper around `sgx_report_t`

nick-mobilecoin:

PR #111, opened 2022-08-29
Add `Key128bit` to wrap `sgx_key_128bit_t`

nick-mobilecoin:

PR #110, opened 2022-08-29
Add `TargetInfo` wrapper for `sgx_target_info_t`

nick-mobilecoin:

PR #103, opened 2022-08-25
Add `ReportBody` wrapper around `sgx_report_body_t`

nick-mobilecoin:

PR #113, opened 2022-08-29
Add `ExtendedAttestationKeyId` rust wrapper

nick-mobilecoin:
Add the `ExtendedAttestationKeyId` to wrap up the `sgx_att_key_id_ext_t`
in a rust structure.

### DIFF
--- a/core/types/src/attestation_key.rs
+++ b/core/types/src/attestation_key.rs
@@ -176,7 +176,7 @@ pub struct ExtendedAttestationKeyId(sgx_att_key_id_ext_t);
 
 impl ExtendedAttestationKeyId {
     // The base attestation key
-    pub fn base_attestation_key(&self) -> QuoteLibAttestationKeyId {
+    pub fn base_key_id(&self) -> QuoteLibAttestationKeyId {
         self.0.base.into()
     }
 
@@ -214,10 +214,7 @@ mod test {
     #[test]
     fn default_extended_attestation_key_id() {
         let key = ExtendedAttestationKeyId::default();
-        assert_eq!(
-            key.base_attestation_key(),
-            QuoteLibAttestationKeyId::default()
-        );
+        assert_eq!(key.base_key_id(), QuoteLibAttestationKeyId::default());
         assert_eq!(key.service_provider_id(), ServiceProviderId([0u8; 16]));
         assert_eq!(key.key_type(), 0);
     }
@@ -233,7 +230,7 @@ mod test {
             reserved: [3u8; 80],
         };
         let key: ExtendedAttestationKeyId = sgx_key.into();
-        assert_eq!(key.base_attestation_key(), base_key);
+        assert_eq!(key.base_key_id(), base_key);
         assert_eq!(key.service_provider_id(), ServiceProviderId([10u8; 16]));
         assert_eq!(key.key_type(), 5);
     }
--- a/core/types/src/attestation_key.rs
+++ b/core/types/src/attestation_key.rs
@@ -2,7 +2,6 @@
 //! Attestation Key types
 
 use crate::{
-    attestation_key::QuoteSignature::{Linkable, UnLinkable},
     new_type_accessors_impls,
     report::{ExtendedProductId, FamilyId},
     ConfigId, FfiError,
@@ -43,22 +42,23 @@ new_type_accessors_impls! {
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 #[non_exhaustive]
 #[repr(u16)]
-enum QuoteSignature {
+pub enum QuoteSignature {
     UnLinkable,
     Linkable,
 }
 
-impl From<sgx_quote_sign_type_t> for QuoteSignature {
-    fn from(sign_type: sgx_quote_sign_type_t) -> QuoteSignature {
+impl TryFrom<sgx_quote_sign_type_t> for QuoteSignature {
+    type Error = FfiError;
+    fn try_from(sign_type: sgx_quote_sign_type_t) -> Result<QuoteSignature, FfiError> {
         match sign_type {
             // Per the header `sgx_quote.h` and
             // https://download.01.org/intel-sgx/sgx-linux/2.17.1/docs/Intel_SGX_Developer_Reference_Linux_2.17.1_Open_Source.pdf
             // the `sgx_att_key_id_ext_t::att_key_type` is only valid for EPID
             // quotes it will be 0 otherwise, which also happens to map to
             // SGX_UNLINKABLE_SIGNATURE.
-            sgx_quote_sign_type_t::SGX_UNLINKABLE_SIGNATURE => UnLinkable,
-            sgx_quote_sign_type_t::SGX_LINKABLE_SIGNATURE => Linkable,
-            v => panic!("Got unknown 'sgx_quote_sign_type_t': {:?}", v),
+            sgx_quote_sign_type_t::SGX_UNLINKABLE_SIGNATURE => Ok(QuoteSignature::UnLinkable),
+            sgx_quote_sign_type_t::SGX_LINKABLE_SIGNATURE => Ok(QuoteSignature::Linkable),
+            v => Err(FfiError::UnknownEnumValue(v.0.into())),
         }
     }
 }
@@ -267,5 +267,17 @@ mod test {
         let mut key = QuoteLibAttestationKeyId::default();
         key.0.mrsigner_length = length;
         assert_eq!(key.mr_signer_key_hash(), Err(FfiError::InvalidInputLength));
+    }
+
+    #[parameterized(
+    unlinkable = {0, Ok(QuoteSignature::UnLinkable)},
+    linkable = {1, Ok(QuoteSignature::Linkable)},
+    out_of_bounds = {2, Err(FfiError::UnknownEnumValue(2))},
+    )]
+    fn try_from_signature_type(raw_value: u32, result: Result<QuoteSignature, FfiError>) {
+        assert_eq!(
+            QuoteSignature::try_from(sgx_quote_sign_type_t(raw_value)),
+            result
+        );
     }
 }
--- a/core/types/src/attestation_key.rs
+++ b/core/types/src/attestation_key.rs
@@ -14,12 +14,35 @@ pub enum MrSignerKeyHash {
     Sha384([u8; 48]),
 }
 
-#[derive(Debug, Default, Clone, Hash, PartialEq, Eq)]
-#[repr(transparent)]
-pub struct AlgorithmId(u32);
+#[derive(Default, Debug, Clone, Hash, PartialEq, Eq)]
+#[non_exhaustive]
+#[repr(u16)]
+pub enum Algorithm {
+    #[default]
+    /// Epid 2.0, Anonymous
+    Epid,
 
-new_type_accessors_impls! {
-    AlgorithmId, u32;
+    /// Reserved
+    Reserved,
+
+    /// ECDSA-256 with P-256 curve, Non Anonymous
+    EcdsaP256,
+
+    /// ECDSA-384 with P-384 curve, Non Anonymous (Not currently supported)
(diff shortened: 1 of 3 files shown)


nick-mobilecoin:
* main
  * **PR #91** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/91" target="_blank">Graphite</a> 
    * **PR #86** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/86" target="_blank">Graphite</a> 
      * **PR #92** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/92" target="_blank">Graphite</a> 
        * **PR #103** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/103" target="_blank">Graphite</a> 
          * **PR #110** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/110" target="_blank">Graphite</a> 
            * **PR #111** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/111" target="_blank">Graphite</a> 
              * **PR #112** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/112" target="_blank">Graphite</a> 
                * **PR #113** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/113" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/113?utm_source=stack-comment).

core/types/src/attestation_key.rs
```diff
@@ -0,0 +1,272 @@
...
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub enum MrSignerKeyHash {
+    Sha256([u8; 32]),
+    Sha384([u8; 48]),
+}
+
+#[derive(Debug, Default, Clone, Hash, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct AlgorithmId(u32);
```
nick-mobilecoin, line 21: The underlying values are available in `sgx_quote3.h`, but that header is in `mc-sgx-dcap-sys-types`  It may be that longer term we move all quote stuff to dcap, not sure.

core/types/src/attestation_key.rs
```diff
@@ -0,0 +1,272 @@
...
+    pub fn version(&self) -> Version {
+        self.0.version.into()
+    }
+
+    /// The hash of the MRSIGNER key.
+    ///
+    /// # Errors
+    /// When the length of the hash is not a valid length for a hash,
+    /// [`sgx_ql_att_key_id_t::mrsigner_length`].
+    pub fn mr_signer_key_hash(&self) -> Result<MrSignerKeyHash, FfiError> {
```
nick-mobilecoin, line 87: Not fond of the asymmetry with this function being fallible, but it seems at some point we have to be able to fail converting from the _loose_ C definition to a more strict rust one.

core/types/src/attestation_key.rs
```diff
@@ -0,0 +1,272 @@
...
+                let mut array: [u8; 48] = [0u8; 48];
+                array.copy_from_slice(&self.0.mrsigner[..48]);
+                Ok(MrSignerKeyHash::Sha384(array))
+            }
+            _ => Err(FfiError::InvalidInputLength),
+        }
+    }
+
+    /// The product ID
+    pub fn product_id(&self) -> ExtendedProductId {
```
nick-mobilecoin, line 104: Intentionally skipped the `prod_id` member since it said it was for legacy.  I think IFF we want it then we add logic here to map into the `ExtendedProductId`.  We could use an enum but that seems like more work for the end user than hopefully necessary.

Also I reused the `ExtendedProductId` from `report.rs`.  They are the same length and I'm pretty sure they're meant to be the same, but for some reason this one here in quote used raw sizes

jcape: We're currently using product ID in all our enclaves, so we will need to expose them both.

jcape (changes requested):
Mostly looks good, but we will need to expose the OG product ID unless it's like never set or something...

core/types/src/attestation_key.rs
nick-mobilecoin: Doing a quick look through the  MobileCoin repo I'm not seeing this product id used.
> prod_id
Legacy Product ID of the QE.
extended_prod_id
Extended Product ID of the QE. All 0's for legacy format enclaves.

I'm not seeing `sgx_ql_att_key_id_t` used anywhere in the MobileCoin repo.

core/types/src/attestation_key.rs
```diff
@@ -0,0 +1,271 @@
...
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub enum MrSignerKeyHash {
+    Sha256([u8; 32]),
+    Sha384([u8; 48]),
+}
+
+#[derive(Debug, Default, Clone, Hash, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct AlgorithmId(u32);
```
awygle, line 20: If this corresponds to `sgx_ql_attestation_algorithm_id_t`, shouldn't we be restricting it from taking on values which are (currently) considered illegal? Currently it's unconstrained (and you even have it with a value of 9 in your test, which is above `SGX_QL_ALG_MAX`)

core/types/src/attestation_key.rs
```diff
@@ -0,0 +1,271 @@
...
+
+new_type_accessors_impls! {
+    Id, u16;
+}
+
+/// The type of quote signature.  Only valid for EPID quotes.
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+#[non_exhaustive]
+#[repr(u16)]
+enum QuoteSignature {
```
awygle, line 46: Is this enum intentionally not public? Should it be public?

core/types/src/attestation_key.rs
```diff
@@ -0,0 +1,271 @@
...
+    fn from(sign_type: sgx_quote_sign_type_t) -> QuoteSignature {
+        match sign_type {
+            // Per the header `sgx_quote.h` and
+            // https://download.01.org/intel-sgx/sgx-linux/2.17.1/docs/Intel_SGX_Developer_Reference_Linux_2.17.1_Open_Source.pdf
+            // the `sgx_att_key_id_ext_t::att_key_type` is only valid for EPID
+            // quotes it will be 0 otherwise, which also happens to map to
+            // SGX_UNLINKABLE_SIGNATURE.
+            sgx_quote_sign_type_t::SGX_UNLINKABLE_SIGNATURE => UnLinkable,
+            sgx_quote_sign_type_t::SGX_LINKABLE_SIGNATURE => Linkable,
+            v => panic!("Got unknown 'sgx_quote_sign_type_t': {:?}", v),
```
awygle, line 61: I'm not sure if a `panic!` here is correct, since because of C/C++ weirdness it's quite possible for a variable of "type `sgx_quote_sign_type_t` to contain a value that's not one of the enumerated values. We might prefer to use `TryFrom` instead.

core/types/src/attestation_key.rs
awygle: Is this something we could catch at the time of constructing a `QuoteLibAttestationKeyId`, instead of later on when we try to access the `SignerKeyHash`?

awygle (changes requested):
I feel that the AlgorithmId struct is too loose currently, I think it should be a non-exhaustive enum. My other comments are more questions than requests for changes.

core/types/src/attestation_key.rs
nick-mobilecoin: converted to `TryFrom`

core/types/src/attestation_key.rs
nick-mobilecoin: My initial thought:
- Re-work `new_type_accessors_impls!` to be `TryFrom` instead of `From`, or perhaps 2 versions for things that can't fail
- For the possible failure types add a trait that has a `validate()` method which returns `Result<Self, FfiError>`
I'm thinking this should probably be a follow on issue, because I bet it will touch a bit of code

awygle: My thought here was that instead of having `struct QuoteLibAttestationKeyId(sgx_quote_lib_attestation_key_id_t)`, we could have something like `struct QuoteLibAttestationKeyId { inner: sgx_quote_lib_attestation_key_id_t }` and then `impl TryFrom<sgx_quote_lib_attestation_key_id_t> for QuoteLibAttestationKeyId { /* ... */ }`. That way we can never construct an invalid `QuoteLibAttestationKeyId`. The tuple struct newtype implicitly allows us to construct a `QuoteLibAttestationKeyId` from _any_ `sgx_quote_lib_attestation_key_id_t`, which is not the semantics we actually want, IMO.

I somewhat agree that this might want to be a follow-on issue, as I suspect this pattern of being able to construct invalid versions of the Rust wrapper types might exist in other places as well.

awygle: Actually on looking into it further it's not the tuple type that causes this but (as you correctly mentioned) the `new_type_accessors_impls` macro, which I don't really like in general as it gives users of the Rust types a tremendous amount of access to the FFI types. We're already using it a lot though and it is very convenient, so... not sure what the right call here is. It would _definitely_ be a follow on issue to make any changes to how we're using that macro.

nick-mobilecoin: >  it gives users of the Rust types a tremendous amount of access to the FFI types
Looking at it, it seems that the `AsMut` is the main thing that seems out of place, if we want to ensure that the rust type is always valid.  I think the rest can stay.
We'll want the `AsRef` to make it easier to pass to the C interfaces:
```rust
let mut rust_version = MyRustType::default();
rust_version.some_setter_method(value);
...
let result = unsafe{ a_pass_by_pointer_c_function(rust_version.as_ref()) }
```
pretty sure the syntax isn't 100%, but hopefully it communicates the idea

core/types/src/attestation_key.rs
nick-mobilecoin: Converted `Algorithm` into an enum, that has a `TryFrom<u32>`.  This should lay the groundwork for when the `mc-sgx-dcap-types` comes into play

core/types/src/attestation_key.rs
nick-mobilecoin: created an enum based on the underlying values, will revisit the home when the `mc-sgx-dcap-types` crate comes about

core/types/src/attestation_key.rs
```diff
@@ -0,0 +1,315 @@
...
+    ServiceProviderId, [u8; 16];
+}
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct ExtendedAttestationKeyId(sgx_att_key_id_ext_t);
+
+impl ExtendedAttestationKeyId {
+    // The base attestation key
+    pub fn base_attestation_key(&self) -> QuoteLibAttestationKeyId {
```
jcape, line 179: Suggestion shortening to `base_key_id()` or just `base()`.

jcape (approved):
LGTM, some things are a bit verbose.

core/types/src/attestation_key.rs
nick-mobilecoin: renamed to `base_key_id`

awygle (approved):
Looks good, thanks!

nick-mobilecoin:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/113?utm_source=stack-comment).


Job Reference: C6ZyZ0Ggec8V0ksW3mYN

nick-mobilecoin:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/113?utm_source=stack-comment).


Job Reference: ANN5tbEFcYrOel5JRpoC

PR #126, opened 2022-09-07
Add `mc-sgx-tservice` crate

nick-mobilecoin:


nick-mobilecoin:
* main
    * **PR #126** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/126" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/126?utm_source=stack-comment).

awygle (approved):

nick-mobilecoin:
* main
    * **PR #126** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/126" target="_blank">Graphite</a>  👈
      * **PR #127** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/127" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/126?utm_source=stack-comment).

nick-mobilecoin:
@awygle requesting a re-review as this felt like more than cosmetic changes to update with #125

tservice/src/seal.rs
```diff
@@ -0,0 +1,92 @@
...
+                mac_pointer,
+                self.data.as_ref().len() as u32,
+                self.data.as_ref().as_ptr(),
+                sealed_data.len() as u32,
+                sealed_data.as_mut_ptr() as *mut sgx_sealed_data_t,
+            )
+        }
+        .into_result()?;
+
+        SealedData::try_from(sealed_data).map_err(|_| Error::Unexpected)
```
nick-mobilecoin, line 70: Not sure if we want a dedicated error type for Sealing/unsealing.
Or perhaps `FfiError` should be expanded to have an `mc_sgx_core_types::Error` variant

tservice/types/src/seal.rs
```diff
@@ -69,6 +77,55 @@ impl<'a> TryFrom<&'a [u8]> for SealedData<'a> {
...
+/// A plain old data type (POD) of the component pieces of the data stored in
+/// [`sgx_sealed_data_t`]
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct UnsealedData<T> {
+    /// The data to be encrypted/sealed
+    pub data: T,
+
+    /// The MAC text that will not be encrypted
+    pub mac: Option<T>,
+}
```
jcape, line 102: This is going to sound weird at first blush, but this should be in the `tservice` crate, not `tservice-types`. I say so because having data that is meant to be sealed be accessible outside the enclave is always an error

jcape (approved):
LGTM, there's a couple docs issues that were annotated, and I'd suggest moving UnsealedData to the enclave, but also: why not `Sealed`/`Unsealed`?

Is "data" not implied ;-)

tservice/types/src/seal.rs
nick-mobilecoin: I think that makes sense I was just blanket throwing all types into the types crates, but as you say `unsealed` data shouldn't leak outside of the enclave, once it does it's sealing it is a false feeling of security.

nick-mobilecoin: done now

tservice/src/seal.rs
```diff
@@ -0,0 +1,112 @@
...
+            )
+        }
+        .into_result()?;
+
+        Sealed::try_from(sealed_data).map_err(|_| Error::Unexpected)
+    }
+
+    /// MAC text to add to the sealed data
+    ///
+    /// The MAC text is also referred to as AAD(Additional Authenticated data)
```
nick-mobilecoin, line 62: I'm for changing this to `aad` or similar, but didn't want to deviate too far from the sgx C API

jcape: I think renaming this to `aad` is actually the way to go, because that's what everyone but Intel calls this. It's worth documenting that this is what Intel calls `mac_text`, but if I see a crypto API that says "AAD" I know what that is instantly, whereas "mac text" could be a weird way of saying "the MAC itself", which we don't care about exposing.

tservice/src/seal.rs
```diff
@@ -0,0 +1,112 @@
...
+use mc_sgx_util::ResultInto;
+
+/// Sealed data builder
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Default)]
+pub struct SealedBuilder<T> {
+    /// The data to be encrypted/sealed
+    data: T,
+
+    /// The MAC text that will not be encrypted
+    mac: Option<T>,
```
jcape, line 19: Aside: this should either be called `aad` or `mac_text`, and the comment should really indicate that they mean the same thing.

jcape (approved):
LGTM, but I'd really like to use "AAD" in the identifiers and document that Intel calls it "mac text".

tservice/src/seal.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// The data to be encrypted/sealed
+    data: T,
+
+    /// The AAD(additional authenticated data) to use in the sealing.
+    /// The Intel docs often refer to this as _MAC text_
+    aad: Option<T>,
+}
+
+impl<T: AsRef<[u8]> + core::default::Default> SealedBuilder<T> {
+    /// A [`SealedBuilder`] from
```
awygle, line 24: From what? :smiley: 
```suggestion
    /// Construct a [`SealedBuilder`] from unsealed data
```

tservice/src/seal.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// * `data` - The data to be encrypted/sealed
+    pub fn new(data: T) -> Self {
+        Self {
+            data,
+            ..Default::default()
+        }
+    }
+
+    /// Build the [`Sealed`] object
+    pub fn build(self) -> Result<Sealed<Vec<u8>>> {
```
awygle, line 36: I think this could probably be non-consuming (`&self` instead of `self`), as well as the other builder methods (`&mut self` instead of `mut self`), since the `sgx_seal_data` call doesn't take ownership of `self.data` or `self.aad`. [The Rust style guidelines recommend using non-consuming builders where possible.](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html)

tservice/src/seal.rs
nick-mobilecoin: renamed to aad

tservice/src/seal.rs
nick-mobilecoin: done

tservice/src/seal.rs
nick-mobilecoin: nice find on the builder guidance, will work to update as suggested
Thanks!

tservice/src/seal.rs
nick-mobilecoin: no complaints so resolved

tservice/src/seal.rs
nick-mobilecoin: doh, 
Thanks!

PR #127, opened 2022-09-07
Add unsealing capability to `mc-sgx-tservice`

nick-mobilecoin:


nick-mobilecoin:
* main
    * **PR #126** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/126" target="_blank">Graphite</a> 
      * **PR #127** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/127" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/127?utm_source=stack-comment).

deny.toml
```diff
@@ -36,8 +36,16 @@ deny = [
     # https://github.com/briansmith/ring/issues/774
     { name = "ring" },
 ]
-skip = []
-skip-tree = []
+skip = [
```
nick-mobilecoin, line 39: Not fond of this, a little surprised there isn't a way to allow path dependencies and only check their transitive dependencies.
This may all go away though if #125 collapses to opaque types as the common test logic would no longer be needed in `mc-sgx-tservice-types`

awygle (approved):

nick-mobilecoin:
* main
    * **PR #126** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/126" target="_blank">Graphite</a> 
      * **PR #127** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/127" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/127?utm_source=stack-comment).

nick-mobilecoin:
* main
    * **PR #126** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/126" target="_blank">Graphite</a> 
      * **PR #127** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/127" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/127?utm_source=stack-comment).

tservice/src/seal.rs
```diff
@@ -71,6 +71,77 @@ impl<T: AsRef<[u8]>> Seal<Vec<u8>> for UnsealedData<T> {
...
+            )
+        };
+        match result {
+            // Per the documentation, UINT32_MAX indicates an error
+            u32::MAX => Err(Error::Unexpected),
+            size => Ok(size as usize),
+        }
+    }
+
+    fn unseal(&self) -> Result<UnsealedData<Vec<u8>>> {
```
jcape, line 112: This necessarily means we're going to need an `Unseal::unseal<T>(&self, &dest: UnsealedData<T>)`, but...

```suggestion
    #[cfg(feature = "alloc")]
    fn unseal_vec(&self) -> Result<UnsealedData<Vec<u8>>> {
```

tservice/src/seal.rs
```diff
@@ -71,6 +71,77 @@ impl<T: AsRef<[u8]>> Seal<Vec<u8>> for UnsealedData<T> {
...
+        let mut data = vec![0; data_length];
+        let mut mac = vec![0; mac_length];
+
+        let mac_pointer = match mac_length {
+            0 => ptr::null_mut(),
+            _ => mac.as_mut_ptr(),
+        };
+
+        let mut data_length = data_length as u32;
+        let mut mac_length = mac_length as u32;
```
jcape, line 124: That these are being passed as `mut` to the FFI binding is giving me pause. I feel like we should return an internal error if `mac_length` or `data_length` has changed...

tservice/test_utils/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2021"
+authors = ["MobileCoin"]
+rust-version = "1.62.1"
+license = "Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/mobilecoinfoundation/sgx"
+description = "Test utilities for the rust wrappers of the `sgx_tservice` library."
+
+categories = ["development-tools::testing"]
+keywords = ["sgx", "test"]
```
jcape, line 15: I'm ambivalent about this... If we're going to go down the "grouping" route I'd like to 1. make more groups and 2. make everything look the same.

jcape (changes requested):
I'd like to chat about the tservice test_utils crate, and whether we should just make it a module in tservice that's feature-disabled by default?

Also we should gate `Unseal::unseal()` (as it currently exists) on "alloc", and either do one that writes to a `dest: UnsealedData<T>`, or maybe is just `From<SealedData<T>> for UnsealedData<T>` :-)

tservice/src/seal.rs
nick-mobilecoin: I think this goes back to the common C api, of here is my buffer and it has room for X bytes, then the function fills the buffer with Y bytes replacing the size with the _used_ size.
```
     *      p_additional_MACtext_length - [IN/OUT] pointer to length of the plaintext data stream in bytes
     *                             If there is not additional data, this parameter should be zero.
```
This may be a failure on my part to [resize](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize) with the values afterward.

Or are you thinking that they shouldn't change since previously we got the needed length?

tservice/src/seal.rs
nick-mobilecoin: I thought alloc  as a feature was only needed for things outside of the enclave.

If it's true we don't need separate alloc logic:
- should `Unsealed` be converted to a `Vec<u8>` instead of a generic?
- should the `Unseal` go away and be methods on the `Unsealed` struct?

tservice/src/seal.rs
nick-mobilecoin: https://github.com/mobilecoinfoundation/sgx/pull/127/commits/18bf2d20d0b8caf3d0dd6c81560ce40336c2faca added logic to error if the sizes differ from what SGX told us originally.

tservice/test_utils/Cargo.toml
nick-mobilecoin: resolving for now as this is in other cargo.toml files and should have a dedicated normalization effort

tservice/src/seal.rs
```diff
@@ -90,10 +90,61 @@ impl<T: AsRef<[u8]> + core::default::Default> SealedBuilder<T> {
     }
 }
 
+pub trait Unseal<T> {
+    /// The length (in bytes) needed to hold the decrypted text
+    fn decrypted_text_len(&self) -> Result<usize>;
+
+    /// Unseal the data in `self`
+    fn unseal(&self) -> Result<T>;
+}
```
jcape, line 99: I'm not sure I like putting the type param on this trait---it feels like it's really only capable of being used for `Vec` and constant arrays.

My preference would be to have something like this:

```rust
pub trait Unseal {
    fn decrypted_text_len(&self) -> Result<usize>;
    fn unseal(&self, dest: &mut [u8]) -> Result<usize>;

    fn unseal_vec(&self) -> Result<Vec<u8>> {
        let mut retval = vec![0u8; self.decrypted_text_len()];
        let retval_len = self.unseal(&mut retval)?;
        retval.truncate(retval_len);
        Ok(retval)
    }
}
```

The common case (give me the data as a `Vec`) is handled, but not at the expense of foreclosing the caller into using the heap or trying to guess at the size of the data from the size of the data in `Sealed<>`

jcape (commented):
I'm ambivalent about Unseal<T>, do you want to chat about it?

deny.toml
nick-mobilecoin: gone now

tservice/src/seal.rs
nick-mobilecoin: Reworked to have an `unseal` and `unseal_to_vec` method.
The trait is generic over `AsRef<[u8]>` which should match the scope of `Sealed`

tservice/src/seal.rs
```diff
@@ -90,10 +90,74 @@ impl<T: AsRef<[u8]> + core::default::Default> SealedBuilder<T> {
...
+                &mut data_length_u32,
+            )
+        }
+        .into_result()?;
+
+        // While the lengths can be modified by `sgx_unseal_data`, we asked the
+        // SGX interface at the top of the function for the required sizes.
+        // If the sizes are different than we have unexpected behavior.
+        if data_length != data_length_u32 as usize || mac_length_u32 != 0 {
+            return Err(Error::Unexpected);
```
awygle, line 139: It may be worth exercising this codepath in the tests.

tservice/src/seal.rs
```diff
@@ -90,10 +90,74 @@ impl<T: AsRef<[u8]> + core::default::Default> SealedBuilder<T> {
...
+    /// Returns the unsealed data into the provided `buffer`.  The returned
+    /// slice will be the exact size of the unsealed data.
+    ///
+    /// # Arguments
+    /// * `buffer` - The buffer to output the unsealed data into.  `buffer`
+    ///   needs to be at least as big as [`decrypted_text_len`].
+    fn unseal<'a>(&self, buffer: &'a mut [u8]) -> Result<&'a mut [u8]> {
+        let data_length = self.decrypted_text_len()?;
+        if buffer.len() < data_length {
+            return Err(Error::InvalidParameter);
```
awygle, line 119: It may be worth exercising this codepath in the tests.

awygle (changes requested):
LGTM, just needs to cover failure cases in Unseal testing.

tservice/src/seal.rs
nick-mobilecoin: Unfortunately this logic we can't exercise without coming up with some kind of framework to run in the enclave.  
If one tries to compile a test for this crate it will fail to link up because it will try to pull in std for the test run, but the enclave code is no_std and when it links in the sgx libs it will be missing the allocator, panic handler, etc.

tservice/src/seal.rs
nick-mobilecoin: see above

tservice/src/seal.rs
awygle: Ah, that's irritating. In that case can you file an issue for future work about figuring out how to execute tests of this type? We're going to run into more of them in the future, I'm sure.

awygle (approved):

tservice/src/seal.rs
nick-mobilecoin: Created, #135

jcape (approved):

PR #130, opened 2022-09-12
Add rust enclave report create and verify wrappers

nick-mobilecoin:
A new trait has been added `mc-sgx-tservice::Report` which provides
functionality to create an `mc-sgx-core-types::Report` and to verify a
`mc-sgx-core-types::Report`.  This functionality operates inside of an
enclave.

PR #131, opened 2022-09-12
Add rust wrapper for sgx_seal_data_ex()

nick-mobilecoin:


nick-mobilecoin:
* main
    * **PR #126** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/126" target="_blank">Graphite</a> 
      * **PR #127** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/127" target="_blank">Graphite</a> 
        * **PR #130** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/130" target="_blank">Graphite</a> 
          * **PR #131** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/131" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/131?utm_source=stack-comment).

tservice/src/seal.rs
```diff
@@ -3,18 +3,32 @@
...
 pub use mc_sgx_tservice_sys::sgx_calc_sealed_data_size;
 pub use mc_sgx_tservice_sys_types::sgx_sealed_data_t;
 pub use mc_sgx_tservice_types::Sealed;
 use mc_sgx_util::ResultInto;
 
+// Default values used in `sgx_seal_data_ex` to behave the same as
+// `sgx_seal_data`. See
+// https://download.01.org/intel-sgx/sgx-linux/2.17.1/docs/Intel_SGX_Developer_Reference_Linux_2.17.1_Open_Source.pdf
+const DEFAULT_MISCELLANEOUS_MASK_FOR_SEAL: u32 = 0xF0000000;
+const DEFAULT_ATTRIBUTES_FLAGS_FOR_SEAL: u64 = 0xFF0000000000000B;
```
awygle, line 17: This is `TSEAL_DEFAULT_FLAGSMASK` in `common/inc/internal/tseal_migration_attr.h`, if you'd prefer to use that.

tservice/src/seal.rs
```diff
@@ -3,18 +3,32 @@
...
 pub use mc_sgx_core_types::{Error, Result};
 pub use mc_sgx_tservice_sys::sgx_calc_sealed_data_size;
 pub use mc_sgx_tservice_sys_types::sgx_sealed_data_t;
 pub use mc_sgx_tservice_types::Sealed;
 use mc_sgx_util::ResultInto;
 
+// Default values used in `sgx_seal_data_ex` to behave the same as
+// `sgx_seal_data`. See
+// https://download.01.org/intel-sgx/sgx-linux/2.17.1/docs/Intel_SGX_Developer_Reference_Linux_2.17.1_Open_Source.pdf
+const DEFAULT_MISCELLANEOUS_MASK_FOR_SEAL: u32 = 0xF0000000;
```
awygle, line 16: This is `TSEAL_DEFAULT_MISCMASK` in `common/inc/internal/tseal_migration_attr.h`, if you'd prefer to use that.

tservice/src/seal.rs
```diff
@@ -68,6 +91,24 @@ impl<T: AsRef<[u8]> + core::default::Default> SealedBuilder<T> {
...
+    pub fn key_policy(mut self, policy: KeyPolicy) -> Self {
+        self.policy = policy;
+        self
+    }
+
+    /// Set the attributes to use for the sealed data
+    ///
+    /// # Arguments
+    /// * `attributes` - The attributes to use when sealing
+    pub fn attributes(mut self, attributes: Attributes) -> Self {
```
awygle, line 107: I'm a little hesitant to expose the `Attributes` "raw" like this. For one thing, sealing will fail if `INIT` and `DEBUG` aren't included in the attribute mask. For another, most of the attribute mask is reserved bits, as are most of the xfrm bits (one could imagine implementing this in the `Attributes` struct's methods, having them take a bitflags struct instead or something). Finally, Intel's guide pretty strongly recommends not moving off of the defaults for this, and I was not able to figure out what this field actually _does_ in the key generation process despite diving pretty deep.

Given the above, and the fact that this method is only called with defaults in the mobilecoin repo, I recommend removing this function entirely.

awygle (changes requested):

tservice/src/seal.rs
nick-mobilecoin: Do you mean directly from the header, or using the name for similarity between the code bases?
- The header is _internal_ implied to be private, so don't think we should use it.
- These are local consts so could name them similarly.  I think for now I'll add a comment referencing the internal name.  If one does feel the names are too long I could understand that.  I usually fall into the verbose over abbreviated camp, but know that it can be taken too far at times.

tservice/src/seal.rs
```diff
@@ -28,7 +42,10 @@ impl<T: AsRef<[u8]> + core::default::Default> SealedBuilder<T> {
     pub fn new(data: T) -> Self {
         Self {
             data,
-            ..Default::default()
+            policy: DEFAULT_KEY_POLICY_FOR_SEAL,
```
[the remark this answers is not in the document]
nick-mobilecoin: I'm reluctant to default `KeyPolicy` to `MRSIGNER` everywhere else.  It's used in the `KeyRequest` and I wouldn't want people to assume that `MRSIGNER` is the preferred key method.  It looks like `sgx_get_key()` isn't currently used by mobilecoin repo, however since this crate is meant to be more generic I think it's better to avoid the default based on only the mobilecoin repo needs.

tservice/src/seal.rs
nick-mobilecoin: Good catch.  Looking further it doesn't appear that `sgx_seal_data_ex` is actually used for the `_ex` part in the mobilecoin repo.  I think the key policy expansion is "ok", but am also fine closing this PR if we think it's not worth doing at this time.

awygle (approved):

nick-mobilecoin:
Thinking about `sgx_seal_data` vs `sgx_seal_data_ex` some more.
As currently proposed, the rust wrappers are only exposing the  `sgx_seal_data_ex` and defaulting to behaving like `sgx_seal_data`.  The defaults are chosen based on the documentation of [`sgx_sealed_data_ex`](https://download.01.org/intel-sgx/sgx-linux/2.17.1/docs/Intel_SGX_Developer_Reference_Linux_2.17.1_Open_Source.pdf#%5B%7B%22num%22%3A331%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C733.5%2C0%5D)

A concern with this is that the defaults are only documented in the pdf or accessible via a private header.  There are no build time assurances that the defaults in the rust interface match those of the C interface.

Looking at the [`sgx_sealed_datat_t`](https://download.01.org/intel-sgx/sgx-linux/2.17.1/docs/Intel_SGX_Developer_Reference_Linux_2.17.1_Open_Source.pdf#%5B%7B%22num%22%3A480%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C364.5%2C0%5D) it contains [`sgx_key_request_t`](https://download.01.org/intel-sgx/sgx-linux/2.17.1/docs/Intel_SGX_Developer_Reference_Linux_2.17.1_Open_Source.pdf#%5B%7B%22num%22%3A472%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C297.75%2C0%5D).  The `sgx_key_request_t` contains the attributes and mask that were used to create the sealed data, and to unseal the data.

We could make a test that seals data via `sgx_seal_data` and then verify the attributes and misc mask in that sealed data match the rust interface defaults.

It's possible the attributes and misc mask have some bits masked off. 
This doesn't seem likely, while some bits may be specific to sealing it would make it difficult for one to take sealed data and reseal an updated version consistently.

jcape (approved):

PR #132, opened 2022-09-12
Idiomatic Rust wrappers for urts types

awygle:
I was unable to get a useful PCL test environment just because it's a complex system so I'm leaving that as future work here.


urts/src/lib.rs
```diff
@@ -93,35 +179,77 @@ impl EnclaveBuilder {
...
+
+impl Enclave {
+    /// Returns the target info for the enclave.
+    pub fn get_target_info(&self) -> Result<TargetInfo, Error> {
+        let mut target_info = MaybeUninit::uninit();
+        unsafe { sgx_get_target_info(self.id, target_info.as_mut_ptr()) }
+            .into_result()
+            .map_err(Error::from)
+            .map(|_| unsafe { target_info.assume_init() }.into())
+    }
```
nick-mobilecoin, line 247: :wrench: consider the following
```suggestion
    pub fn target_info(&self) -> Result<TargetInfo, Error> {
        let mut target_info = sgx_target_info_t::default();
        unsafe { sgx_get_target_info(self.id, &mut target_info) }
            .into_result()?;
        Ok(target_info.into())
    }
```
The main advantages to this is that one ensures `target_info` is initialized.  Yes it's all zeros and they may not be `valid` values if `sgx_get_target_info()` was never called, but if for some reason `sgx_get_target_info()` fails to initialize a field, *always* pre-initializing to zero will provide consistent behaviour that can be debugged versus whatever happens to be on the stack uninitialized values.

urts/src/lib.rs
```diff
@@ -93,35 +179,77 @@ impl EnclaveBuilder {
...
+            debug: false,
+            pcl_key: None,
+            kss_config: None,
+        })
+    }
+}
+
+impl Enclave {
+    /// Returns the target info for the enclave.
+    pub fn get_target_info(&self) -> Result<TargetInfo, Error> {
```
nick-mobilecoin, line 241: :wrench: rust lang discourages `get_` https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter
This is rust-lang, but most of our conventions usually mimic them.

urts/src/lib.rs
```diff
@@ -93,35 +179,77 @@ impl EnclaveBuilder {
...
+        unsafe { sgx_get_target_info(self.id, target_info.as_mut_ptr()) }
+            .into_result()
+            .map_err(Error::from)
+            .map(|_| unsafe { target_info.assume_init() }.into())
+    }
+
+    /// Returns a reference to the enclave ID.
+    /// Returns by reference because enclave ID will not be valid after the
+    /// enclave is dropped.
+    pub fn get_id(&self) -> &sgx_enclave_id_t {
```
nick-mobilecoin, line 252: :wrench: :wrench: rust lang discourages `get_` https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter
This is rust-lang, but most of our conventions usually mimic them.
```suggestion
    pub fn id(&self) -> &sgx_enclave_id_t {
```

urts/src/lib.rs
```diff
@@ -143,54 +271,102 @@ impl Drop for Enclave {
...
+
     #[test]
     fn calling_into_an_enclave_function_provides_valid_results() {
         // Note: the `debug()` was added to ensure proper builder behavior of
         // the `create()` method.  It could go away if another test has need
         // of similar behavior.
-        let enclave = EnclaveBuilder::new(ENCLAVE).debug(true).create().unwrap();
+        let enclave = EnclaveBuilder::from(ENCLAVE).debug(true).create().unwrap();
+        let id = enclave.get_id();
+        println!("{:X}", id);
```
nick-mobilecoin, line 342: :wrench: looks like this is debug code that should probably go away
```suggestion
```

nick-mobilecoin (approved):
mainly suggestions and questions about API

test_enclave/src/pcl_key.bin
```diff
@@ -0,0 +1 @@
+""33DDUUffww����
```
jcape, line 1: We need a `.gitattributes` file to ensure these types of files aren't treated as text (and potentially munged due to CRLF foolishness). I'd recommend using [the example from mobilecoin.git](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/.gitattributes). This is kind of pulling a thread, because it will require [defining the merge driver as well](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/.gitconfig).

urts/src/lib.rs
```diff
@@ -2,22 +2,57 @@
...
+use mc_sgx_urts_sys_types::{sgx_enclave_id_t, sgx_kss_config_t};
+use mc_sgx_util::ResultInto;
+use std::{ffi::c_void, fs::File, io::Read, mem::MaybeUninit, os::raw::c_int, path::Path, ptr};
 
-#[derive(Debug, PartialEq, Eq)]
-pub enum Error {
-    // An error provided from the SGX SDK
-    SgxStatus(sgx_status_t),
+/// Structure defining configuration for Key Sharing and Separation
+pub struct KssConfig {
```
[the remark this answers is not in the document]
jcape: I think we can keep `KssConfig` here, unless there's some use for it inside an enclave.

That said, this struct does need to be changed to wrap the bindgen type, because otherwise we have to keep them in sync.

urts/src/lib.rs
```diff
@@ -2,22 +2,57 @@
...
+}
+
+impl From<sgx_kss_config_t> for KssConfig {
+    fn from(input: sgx_kss_config_t) -> KssConfig {
+        KssConfig {
+            config_id: input.config_id,
+            config_svn: input.config_svn,
+        }
+    }
 }
```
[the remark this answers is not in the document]
jcape: I think we should do the newtype wrapper and `KssConfig::new(config_id: ConfigId, config_svn: ConfigSvn)`, under the assumption that there isn't a particular use for mutable accessors...

urts/src/lib.rs
```diff
@@ -32,20 +67,24 @@ pub struct EnclaveBuilder {
...
-    /// * `bytes` - The bytes representing the enclave file.  This should be a
-    ///   signed enclave.
-    pub fn new(bytes: &[u8]) -> EnclaveBuilder {
-        EnclaveBuilder {
-            bytes: bytes.into(),
-            debug: false,
-        }
+    /// * `path` - The path to the enclave file. This should be a signed
+    ///   enclave.
+    pub fn new(path: impl AsRef<Path>) -> std::io::Result<EnclaveBuilder> {
```
jcape, line 85: Recommend just deleting this and leaving `TryFrom<T> where T: AsRef<Path>` be the way. Adding a usage example on `EnclaveBuilder` for this and the from bytes impls is going to be superior than having multiple paths to the same functionality anyways...

urts/src/lib.rs
```diff
@@ -59,18 +98,65 @@ impl EnclaveBuilder {
         self
     }
 
+    /// Enable Intel's Protected Code Loader for the enclave
+    ///
+    /// # Arguments
+    ///
+    /// * `key` - The sealed PCL key to use for loading the enclave
+    #[must_use]
+    pub fn pcl(mut self, key: Vec<u8>) -> EnclaveBuilder {
```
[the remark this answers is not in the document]
jcape: I don't think it's actually useful to have this be `Option<>`---`EnclaveBuilder` should only live for as long as it takes to start an enclave anyways.

urts/src/lib.rs
```diff
@@ -59,18 +98,65 @@ impl EnclaveBuilder {
...
+        self
+    }
+
+    /// Disable Intel's Protected Code Loader for the enclave
+    #[must_use]
+    pub fn no_pcl(mut self) -> EnclaveBuilder {
+        self.pcl_key = None;
+        self
+    }
+
```
jcape, line 118: This is overkill...

```suggestion
```

urts/src/lib.rs
```diff
@@ -59,18 +98,65 @@ impl EnclaveBuilder {
...
+    /// Disable Key Separation & Sharing for the enclave
+    #[must_use]
+    pub fn no_kss(mut self) -> EnclaveBuilder {
+        self.kss_config = None;
+        self
+    }
+
```
jcape, line 136:
```suggestion
```

urts/src/lib.rs
```diff
@@ -93,35 +179,77 @@ impl EnclaveBuilder {
...
+        .map(|_| Enclave { id: enclave_id })
     }
 }
 
 impl From<Vec<u8>> for EnclaveBuilder {
-    fn from(bytes: Vec<u8>) -> EnclaveBuilder {
+    fn from(input: Vec<u8>) -> EnclaveBuilder {
         EnclaveBuilder {
-            bytes,
+            bytes: input,
```
jcape, line 195: We can just leave this as bytes :-)

urts/src/lib.rs
```diff
@@ -93,35 +179,77 @@ impl EnclaveBuilder {
...
+            pcl_key: None,
+            kss_config: None,
+        }
     }
 }
 
-impl Deref for Enclave {
-    type Target = sgx_enclave_id_t;
-    fn deref(&self) -> &Self::Target {
+impl TryFrom<File> for EnclaveBuilder {
```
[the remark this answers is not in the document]
jcape: I think `impl<P: AsRef<Path>> TryFrom<&'a P> for EnclaveBuilder` may be the way to do this.

urts/src/lib.rs
jcape: Yep, we should use `target_info()`

jcape (changes requested):

test_enclave/build.rs
```diff
@@ -149,7 +178,10 @@ where
 ///
```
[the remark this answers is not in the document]
awygle: Ah I see what you mean, I should pass `keyfile.is_some()` rather than the whole `keyfile` since its value isn't used. Fair enough.

I am hesitant to link the PCL library when it's not used... just because I haven't dug into it very deeply so I don't really know whether it would cause some issue down the line.

test_enclave/build.rs
```diff
@@ -177,7 +209,17 @@ fn build_dynamic_enclave_binary<P: AsRef<Path>>(static_enclave: P) -> PathBuf {
...
-        .args(&["--whole-archive", &trts, "--no-whole-archive"])
+        .arg("--whole-archive")
+        .arg(&trts);
+    if let Some(_) = keyfile {
+        if cfg!(feature = "sim") {
+            command.arg("-lsgx_pclsim");
+        } else {
+            command.arg("-lsgx_pcl");
+        }
+    }
```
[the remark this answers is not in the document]
awygle: Did this get added in a later commit? I think I just didn't know it existed when I wrote this piece of code...

test_enclave/src/config_pcl.xml
```diff
@@ -0,0 +1,15 @@
+<EnclaveConfiguration>
```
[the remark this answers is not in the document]
awygle: Hm, I suppose not. I'll remove it.

test_enclave/src/pcl_key.bin
awygle: Git is so elegant and well designed!

Gotcha, I'll add this...

urts/src/lib.rs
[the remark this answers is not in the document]
awygle: OK, I'll do that.

For posterity - the reason I did this originally was because I figured there _was_ a use for mutable accessors, and the idiomatic Rust way to do that is just to make the fields public rather than mess around with `set_config(&mut self, config: ConfigId)` functions. James' comment about keeping them in sync is a better argument for the newtype though.

urts/src/lib.rs
awygle: Just to make sure I'm following: you want the only way to construct this type to be by using a `TryFrom` or a `From` from another type?

I can do that, I just don't love it from an expressiveness standpoint. I do get your point about having multiple paths to the same functionality, and I guess we do have this kind of `From`-first approach in a lot of other cases, but to me `EnclaveBuilder::new(thing)` is always going to be clearer in intent than `thing.from()`.

urts/src/lib.rs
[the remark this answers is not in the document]
awygle: I think my preferred solution here would be `.debug()` and `.no_debug()`, if we're worried about consistency.

James, I don't really understand your comment: are you saying get rid of the `no_kss` and `no_pcl` methods? It sounds like you're suggesting changing the `pcl_key: Option<Vec<u8>>` to be `pcl_key: Vec<u8>`, but I don't see how that would make sense given that we're certainly not going to force everyone to use PCL.

urts/src/lib.rs
```diff
@@ -59,18 +98,65 @@ impl EnclaveBuilder {
...
+    /// ID.
     pub fn create(mut self) -> Result<Enclave, Error> {
         let mut enclave_id: sgx_enclave_id_t = 0;
-        let result = unsafe {
+        let mut ex_features = 0;
+        let mut ex_features_p: [*const c_void; 32] = [ptr::null(); 32];
+
+        if let Some(pcl_key) = self.pcl_key {
+            ex_features |= SGX_CREATE_ENCLAVE_EX_PCL;
+            ex_features_p[SGX_CREATE_ENCLAVE_EX_PCL_BIT_IDX as usize] =
```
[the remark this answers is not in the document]
awygle: Again, I think that functionality wasn't in the version of `mc-sgx-core-build` I actually wrote this against and I didn't notice it after I rebased.

urts/src/lib.rs
awygle: I don't... remember why I did this. :shrug: I'll revert it.

urts/src/lib.rs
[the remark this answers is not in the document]
awygle: I can drop the `TryFrom` for `File`. It's in there because I hate being in a situation where I have a `Path` or a `File` but the API I'm talking to requires me to only use the other one. One concrete example of this is when testing using temporary files, it's often much more annoying (for valid security-type reasons) to get a `NamedTempFile` than to get a `TempFile`.

urts/src/lib.rs
awygle: :+1:. I think I just glazed over and copied the SGX name directly.

urts/src/lib.rs
awygle: :+1:

urts/src/lib.rs
awygle: Oops.

test_enclave/build.rs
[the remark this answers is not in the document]
nick-mobilecoin: It was there for initial ffi release, https://docs.rs/mc-sgx-core-build/0.1.0/mc_sgx_core_build/fn.sgx_library_suffix.html but it could probably use better documentation

urts/src/lib.rs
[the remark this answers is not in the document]
nick-mobilecoin: pretty sure this one is newer, as I bet `as usize` was added due to compilation errors

urts/src/lib.rs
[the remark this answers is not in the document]
jcape: Ah, yeah, sorry, I was responding directly to Nick's suggested function prototype:
```rust
pub fn pcl(mut self, key: Option<Vec<u8>>) -> EnclaveBuilder;
```

urts/src/lib.rs
awygle: Actually, you can't do a `TryFrom<T> where T: AsRef<Path>`, because of the blanket `TryFrom` impl in `core`. So I'm leaving `new` in place.

urts/src/lib.rs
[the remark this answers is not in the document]
awygle: Per discussion with James, keeping `TryFrom<File>`, and as said above, can't do `TryFrom` with generics so leaving it in `new`.

urts/src/lib.rs
```diff
@@ -93,35 +179,77 @@ impl EnclaveBuilder {
...
+            pcl_key: None,
+            kss_config: None,
         }
     }
 }
 
-impl Enclave {
-    fn new(id: sgx_enclave_id_t) -> Enclave {
-        Enclave { id }
+impl From<&[u8]> for EnclaveBuilder {
```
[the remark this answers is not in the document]
awygle: I think keeping `TryFrom<File>` is worth this small repetition. I did reorganize the `From` impls so that they all call the `From<Vec<u8>>`, since everything needs to be turned into a vec at the bottom anyway.

urts/sys/types/src/lib.rs
```diff
@@ -7,3 +7,12 @@ use mc_sgx_core_sys_types::{
     sgx_attributes_t, sgx_config_id_t, sgx_config_svn_t, sgx_misc_select_t,
 };
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+
+impl Default for sgx_kss_config_t {
+    fn default() -> sgx_kss_config_t {
+        sgx_kss_config_t {
+            config_id: [0; 64],
```
nick-mobilecoin, line 14: :wrench: Looks like I used a `64` or 2 in `mc-sgx-core-sys-types` :(
Although the rust compiler should bark if that `64` ever changes
```suggestion
    sgx_attributes_t, sgx_config_id_t, sgx_config_svn_t, sgx_misc_select_t, SGX_CONFIGID_SIZE,
};
include!(concat!(env!("OUT_DIR"), "/bindings.rs"));

impl Default for sgx_kss_config_t {
    fn default() -> sgx_kss_config_t {
        sgx_kss_config_t {
            config_id: [0; SGX_CONFIGID_SIZE],
```

nick-mobilecoin (approved):
LGTM

jcape (approved):

PR #137, opened 2022-09-19
Add `mc-sgx-dcap_quoteverify` crate

nick-mobilecoin:


nick-mobilecoin:
* main
  * **PR #128** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/128" target="_blank">Graphite</a> 
    * **PR #137** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/137" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/137?utm_source=stack-comment).

dcap/quoteverify/src/lib.rs
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) 2022 MobileCoin Foundation
+
+#![doc = include_str!("../README.md")]
+#![deny(missing_docs, missing_debug_implementations)]
+
+use mc_sgx_dcap_quoteverify_types::Path as QvPath;
```
nick-mobilecoin, line 6: Started this in the `lib.rs` module for now, I'm up for suggestions on name for a containing module

dcap/quoteverify/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+///
+/// Returns [`Quote3Error`] when
+/// * `path` does not point to a file
+/// * `path` is longer than 259 (bytes)
+/// * `path` contains a null (0) byte.
+///
+/// # Arguments
+/// * `path_type` - Which path to set
+/// * `path` - The path value to use.  This is the full path to a file.
+pub fn set_path<P: AsRef<Path>>(path_type: QvPath, path: P) -> Result<(), Quote3Error> {
```
awygle, line 25: Hm. Is there any Rust idiomatic guidance on global configuration methods like this? It feels a bit sketchy as it's modifying global state, but doing it secretly. I am wondering if we should make some kind of Qv struct to attach this method and the follow-on `load_policy` and `supplemental_data_size` to. That would allow us to mark it as `!Send` (actually I think that happens automatically) and force people to think before pulling the rug out from under another thread.

awygle (approved):
I think I'm OK with this as it seems the common Rust thing to do is not to worry about the sort of issues I mentioned but we should at least discuss the options before committing too deeply.

jcape (approved):
I think we really want to fix this, because it _really_ feels like we're leaving a footgun for future users... That said, it doesn't need to get fixed in this PR.

dcap/quoteverify/src/lib.rs
nick-mobilecoin: created #145

nick-mobilecoin:
**Graphite Merge Job**

Current status: 🔄 Waiting for checks to pass

This pull request is currently waiting for CI and mergeability checks to pass before being merged.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/137?utm_source=stack-comment).


Job Reference: 6Shi6vPuuE79lf4JMQAu

PR #165, opened 2022-10-11
Add validation of `data` and `aad` values

nick-mobilecoin:
Previously the `data` and `aad` values to the
`mc-sgx-tservicer::SealedBuilder` were only checked by the SGX SDK.
This resulted in a generic `InvalidParamter` error that had no context.
Now the `data` and `aad` values are ensured to be valid when they are
provided to the `SealedBuilder`.

PR #164, opened 2022-10-11
Add specific errors for sealing failures

nick-mobilecoin:
Add errors that provide more information on why a failure to seal data
occurs.

PR #163, opened 2022-10-10
Only support `sim` in test_enclave

nick-mobilecoin:
Previously the `test_enclave` had a `sim` feature.  The `sim` feature
was used to build the `test_enclave` in a sim environment.  Cargo
currently doesn't support feature propagation to _only_ dev
dependencies. Always propagating `sim` to the `test_enclave` resulted in
failures to package `mc-sgx-urts` as `mc-sgx-urts` was now looking for a
published version of `test_enclave`.  Now `test_enclave` will only build
for the simulation SGX SDK, there is no feature flag to add or remove
this.

PR #159, opened 2022-10-10
(cargo-release) start next development iteration 0.2.1-pre

nick-mobilecoin:

PR #166, opened 2022-10-12
Add `PathInitializer` to `mc-sgx-dcap-ql`

nick-mobilecoin:
Add a `PathInitializer` struct to `mc-sgx-dcap-ql` to ensure that the
necessary DCAP quote library paths are set prior to making quote library
calls.  As well as ensuring that the paths are only initialized once.

### DIFF
--- a/dcap/ql/src/lib.rs
+++ b/dcap/ql/src/lib.rs
@@ -8,5 +8,24 @@ extern crate alloc;
 mod quote3;
 mod quote_enclave;
 
+use mc_sgx_dcap_types::Quote3Error;
 pub use quote3::TryFromReport;
-pub use quote_enclave::{load_policy, set_path, QeTargetInfo};
+pub use quote_enclave::{load_policy, PathInitializer, QeTargetInfo};
+
+/// Errors interacting with quote library functions
+#[derive(Clone, Debug, displaydoc::Display, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[non_exhaustive]
+pub enum Error {
+    /// Paths have already been initialized
+    PathsInitialized,
+    /// Error from SGX quoting library function: {0}
+    Sgx(Quote3Error),
+    /// Failed ot convert a path to a string.  Path {0}
+    PathStringConversion(String),
+}
+
+impl From<Quote3Error> for Error {
+    fn from(src: Quote3Error) -> Self {
+        Self::Sgx(src)
+    }
+}
--- a/dcap/ql/src/quote_enclave.rs
+++ b/dcap/ql/src/quote_enclave.rs
@@ -15,8 +15,8 @@ use mc_sgx_core_types::TargetInfo;
 use mc_sgx_dcap_ql_types::PathKind;
 use mc_sgx_dcap_types::RequestPolicy;
 use mc_sgx_util::ResultInto;
-use once_cell::sync::OnceCell;
-use std::{ffi::CString, os::unix::ffi::OsStrExt, path::Path};
+use once_cell::sync::Lazy;
+use std::{ffi::CString, os::unix::ffi::OsStrExt, path::Path, sync::Mutex};
 
 /// A convenience type alias for a `Result` which contains an [`Error`].
 pub type Result<T> = CoreResult<T, Error>;
@@ -24,14 +24,15 @@ pub type Result<T> = CoreResult<T, Error>;
 /// Initialization of the paths for the quoting enclaves and quote provider
 /// library
 ///
-/// This can only be called once during process start up utilizing
+/// This should only be called once during process start up utilizing
 /// [`PathInitializer::with_paths()`] or [`PathInitializer::try_default()`].
 /// If a consumer of this crate does not explicitly initialize the paths, then
 /// they will be defaulted on the first call to an SGX function that needs the
 /// paths set.
 #[derive(Debug)]
 pub struct PathInitializer;
-static PATH_INITIALIZER: OnceCell<PathInitializer> = OnceCell::new();
+
+static PATH_INITIALIZER: Lazy<Mutex<Option<PathInitializer>>> = Lazy::new(|| Mutex::new(None));
 
 impl PathInitializer {
     /// Try to initialize the paths to the default for the system
@@ -78,19 +79,18 @@ impl PathInitializer {
         quote_provider_library: Option<P>,
         id_enclave: P,
     ) -> Result<()> {
-        let mut paths_set = false;
-        PATH_INITIALIZER.get_or_try_init(|| {
-            paths_set = true;
+        let mut value = PATH_INITIALIZER.lock().expect("Mutex has been poisoned");
+        if value.is_none() {
             Self::set_paths(
                 quoting_enclave,
                 provisioning_certificate_enclave,
                 quote_provider_library,
                 id_enclave,
-            )
-        })?;
-        match paths_set {
-            true => Ok(()),
-            false => Err(Error::PathsInitialized),
+            )?;
+            *value = Some(PathInitializer);
+            Ok(())
+        } else {
+            Err(Error::PathsInitialized)
         }
     }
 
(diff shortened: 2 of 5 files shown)


nick-mobilecoin:
* main
  * **PR #158** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/158" target="_blank">Graphite</a> 
    * **PR #159** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/159" target="_blank">Graphite</a> 
      * **PR #163** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/163" target="_blank">Graphite</a> 
        * **PR #164** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/164" target="_blank">Graphite</a> 
          * **PR #165** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/165" target="_blank">Graphite</a> 
            * **PR #166** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/166" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/166?utm_source=stack-comment).

dcap/ql/src/lib.rs
```diff
@@ -8,5 +8,24 @@ extern crate alloc;
...
+pub use quote_enclave::{load_policy, PathInitializer, QeTargetInfo};
+
+/// Errors interacting with quote library functions
+#[derive(Clone, Debug, displaydoc::Display, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[non_exhaustive]
+pub enum Error {
+    /// Paths have already been initialized
+    PathsInitialized,
+    /// Error from SGX quoting library function: {0}
+    Sgx(Quote3Error),
```
nick-mobilecoin, line 22: I've added a task to #145 to further expand this error. I think it's best for #164 to get settled before investing too much here, also this PR already got bigger than I wanted :(.

dcap/ql/src/quote_enclave.rs
```diff
@@ -62,10 +193,36 @@ mod test {
...
-    use std::fs;
+    use serial_test::serial;
+    use std::{ffi::c_void, fs};
     use tempfile::tempdir;
     use yare::parameterized;
 
+    /// Resets the [`PATH_INITIALIZER`] to being uninitialized.  This is *not*
+    /// thread safe so tests that make use of this need to utilize the
+    /// `#[serial]` macro.
+    fn reset_path_initializer() {
```
nick-mobilecoin, line 204: I do not like this, but I would rather have unsafe test code than untested normal code.
Especially if we want to support retrying path initialization, I think we'll want tests
From task list on #145
>  Consider tests and logic to allow reattempts at initializing the paths if they fail to initialize.

dcap/ql/src/quote_enclave.rs
```diff
@@ -8,17 +8,167 @@
...
+/// library
+///
+/// This can only be called once during process start up utilizing
+/// [`PathInitializer::with_paths()`] or [`PathInitializer::try_default()`].
+/// If a consumer of this crate does not explicitly initialize the paths, then
+/// they will be defaulted on the first call to an SGX function that needs the
+/// paths set.
+#[derive(Debug)]
+pub struct PathInitializer;
+static PATH_INITIALIZER: OnceCell<PathInitializer> = OnceCell::new();
```
nick-mobilecoin, line 34: One idea that come up in talking with @remoun is we could pass the paths as a plain old data type to the functions that need them.  And set them in those function prior to calling the other sgx methods.  

This wouldn't account for threading, but would avoid the global singleton.

NotGyro (approved):
This looks good.

dcap/ql/src/quote_enclave.rs
nick-mobilecoin: Converted to a lazy mutex

dcap/ql/src/quote_enclave.rs
nick-mobilecoin: offline discussion seemed to want to stick with a singleton style

dcap/ql/src/quote_enclave.rs
```diff
@@ -8,17 +8,167 @@
...
+/// [`PathInitializer::with_paths()`] or [`PathInitializer::try_default()`].
+/// If a consumer of this crate does not explicitly initialize the paths, then
+/// they will be defaulted on the first call to an SGX function that needs the
+/// paths set.
+#[derive(Debug)]
+pub struct PathInitializer;
+
+static PATH_INITIALIZER: Lazy<Mutex<Option<PathInitializer>>> = Lazy::new(|| Mutex::new(None));
+
+impl PathInitializer {
```
nick-mobilecoin, line 37: @jcape in discussions yesterday, you mentioned an inner type.  I don't think it's necessary, bringing it up in case I missed something and to communicate that I didn't ignore or forget the comment.

dcap/ql/src/quote_enclave.rs
```diff
@@ -8,17 +8,167 @@
...
+    ///   converted to a [`CString`]
+    /// * [`Error::Sgx`] if
+    ///     * one of the paths does not point to a file
+    ///     * one of the paths is longer than 259 (bytes)
+    ///     * one of the paths contains a null (0) byte.
+    pub fn with_paths<P: AsRef<Path>>(
+        quoting_enclave: P,
+        provisioning_certificate_enclave: P,
+        quote_provider_library: Option<P>,
+        id_enclave: P,
```
remoun, line 80: Consider moving these 4 fields into a holder struct
```suggestion
        paths: EnclavePaths,
```
with
```rust
struct EnclavePaths {
    quoting_enclave: PathBuf,
    provisioning_certificate_enclave: PathBuf,
    quote_provider_library: Option<PathBuf>,
    id_enclave: PathBuf,
}

Comment (jcape) [reply dcap/ql/src/quote_enclave.rs:80]: I disagree with this suggestion, but what should change about this is using `P1, P2, P3, P3`. Specifically, I'm guessing that if you tried to set paths from a mix of `String`, `&str`, `&Path`, and `PathBuf` types you'd get failures because they aren't all the same `P` type.

Comment (nick-mobilecoin) [reply dcap/ql/src/quote_enclave.rs:80]: >  disagree with this suggestion, but what should change about this is using P1, P2, P3, P3
To clarify @james you *disagree* with the suggestion, 
but you're saying the current implementation should be using 4 different values to allow mixing asref path types.

Curious why do you disagree? 
One advantage of the container is it can prevent transposing `with_paths(pce, qe, Some(id), qpl)`

Comment (awygle) [reply dcap/ql/src/quote_enclave.rs:80]: @jcape is right that you need four type variables in the current implementation.

I'm ambivalent about @remoun's suggestion... it is pretty common in Rust but it really only helps if you want the moral equivalent of default arguments or something, IMO. I think I'm a weak "no" on `EnclavePaths`.

Comment (remoun) [reply dcap/ql/src/quote_enclave.rs:80]: To clarify, I meant a named struct would provide a compile-time check that the paths are passed in correctly.

Comment (nick-mobilecoin) [reply dcap/ql/src/quote_enclave.rs]: Updated to use a type per path

Comment (jcape) [reply dcap/ql/src/quote_enclave.rs]: Yeah, the difference is really between these two calling patterns:

```rust
func(path1, path2, path3, path4);
func_s(Paths { path1, path2, path3, path4 });
```

There is an added margin of safety that's added here, in it does require that your calling code uses the same names as the function args (which it may not):

```rust
func_s(Paths {
    path1: path2,
    path2: path1,
    path3,
    path4,
});
```

remoun: Yes, the extra typing can be trivially bypassed, but it would make it harder to *accidentally* swap the arguments, by making such bugs apparent at the call site, e.g. `foo_path: bar_path`

dcap/ql/src/quote_enclave.rs
```diff
@@ -8,17 +8,179 @@
...
+        quoting_enclave: P1,
+        provisioning_certificate_enclave: P2,
+        quote_provider_library: Option<P3>,
+        id_enclave: P4,
+    ) -> Result<()>
+    where
+        P1: AsRef<Path>,
+        P2: AsRef<Path>,
+        P3: AsRef<Path>,
+        P4: AsRef<Path>,
```
remoun, line 86: Optional/Equivalent: Replace some of the generic types with `impl AsRef<Path>`
```suggestion
    pub fn with_paths<P: AsRef<Path> = &Path>(
        quoting_enclave: impl AsRef<Path>,
        provisioning_certificate_enclave: impl AsRef<Path>,
        quote_provider_library: Option<P>,
        id_enclave: impl AsRef<Path>,
    ) -> Result<()>
```

I'm not 100% sure about `P: AsRef<Path> = &Path`; the idea is to allow passing `None` without naming a type

remoun (approved):
LGTM

dcap/ql/src/quote_enclave.rs
```diff
@@ -8,17 +8,179 @@
...
+        quoting_enclave: P1,
+        provisioning_certificate_enclave: P2,
+        quote_provider_library: Option<P3>,
+        id_enclave: P4,
+    ) -> Result<PathInitializer>
+    where
+        P1: AsRef<Path>,
+        P2: AsRef<Path>,
+        P3: AsRef<Path>,
+        P4: AsRef<Path>,
```
jcape, line 129: 👍🏻

jcape (approved):

dcap/ql/src/quote_enclave.rs
nick-mobilecoin: Regarding the `P: AsRef<Path> = &Path`, https://github.com/rust-lang/rust/issues/36887
It looks like it's not meant to work on functions :(

dcap/ql/src/quote_enclave.rs
nick-mobilecoin: no barks resolving

PR #179, opened 2022-10-24
Ensure types crates don't use alloc with CI

nick-mobilecoin:
Update the crates to build with the `thumbv7m-none-eabi` target using
only the `core` crate from std

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -244,6 +244,33 @@ jobs:
         with:
           files: lcov.info
 
+  build-no-alloc:
+    runs-on: ubuntu-20.04
+    needs:
+      - "rustfmt"
+      - "markdown-lint"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/sgxsdk
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          override: true
+      - run: rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
+      - uses: r7kamura/rust-problem-matchers@v1
+      - name: Build types with no alloc crate
+        # Some notes on this build command:
+        # - The vendored headers are used to get the necessary DCAP headers
+        # - The installed `tlibc` is used to get a compilable `time.h` for the target.
+        # - In the unlikely event that `thumbv7m-none-eabi` was installed with rustup, this would error out with
+        #   duplicate core symbols due to `-Z build-std=core`.
+        run: |
+          cargo metadata --no-deps --format-version=1 |  \
+            jq -r '.packages[].name' | \
+            grep -e types | \
+            xargs -n1 sh -c 'CFLAGS="-isystem${GITHUB_WORKSPACE}/core/build/headers -isystem/opt/intel/sgxsdk/include/tlibc" cargo build -Z build-std=core --target thumbv7m-none-eabi -p $0 || exit 255'
+
   notify:
     runs-on: ubuntu-latest
     if: failure() && ${{ github.event_name }} == 'push'
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -251,7 +251,6 @@ jobs:
       - "markdown-lint"
     steps:
       - uses: actions/checkout@v3
-      - uses: ./.github/actions/sgxsdk
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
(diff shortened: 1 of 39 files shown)


nick-mobilecoin:
* main
  * **PR #179** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/179" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/179?utm_source=stack-comment).

core/build/src/lib.rs
```diff
@@ -69,13 +69,30 @@ pub fn sgx_builder() -> Builder {
...
     cargo_emit::rerun_if_changed!(include_path);
 
     builder
 }
 
+// Gets the `CFLAGS` from the environment, if any.  When there are no `CLFAGS`
+// will return an empty vector.
+// The `CFLAGS` will be split on whitespace in order to allow for multiple
+// arguments. This does *not* attempt to handle escaped shell characters,
+// https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html
```
nick-mobilecoin, line 84: Up for suggestions on a better way to handle getting system headers to custom target triples, in particular `time.h` for `time_t`. 

I did find https://docs.rs/shlex/latest/shlex/ but didn't want to invest too much in this `CFLAGS` path if we think there may be a better way.

Cargo.toml
```diff
@@ -37,6 +37,14 @@ exclude = [
     "test_enclave",
 ]
 
+# We need to explicitly specify resolver 2.
```
nick-mobilecoin, line 40: The cake is a lie

I'm hoping someone might be able to prove me wrong here, but it seems to break without this.

core/types/src/lib.rs
```diff
@@ -23,11 +23,27 @@ pub use crate::{
...
+// For targets that don't have a random number source we force it to always
+// fail.
+// Per https://docs.rs/getrandom/latest/getrandom/macro.register_custom_getrandom.html
+// this function will *only* be used if getrandom doesn't know of a native
+// secure spng
+#[cfg(target_os = "none")]
+use getrandom::register_custom_getrandom;
+
+#[cfg(target_os = "none")]
+register_custom_getrandom!(always_fail);
```
nick-mobilecoin, line 44: This macro choked on the crate specific `Result`, since there was only one use of it, I removed it.

Cargo.toml
awygle: The relevant Cargo issue here is https://github.com/rust-lang/cargo/issues/4866 (and yes, `resolver = "2"` is required).

awygle (approved):
I think this is basically fine. It would be nice to have a more elegant solution but that doesn't seem to be possible given the current state of tooling.

core/build/src/lib.rs
jcape: I think `shlex` may be a useful addition to `mc-util-build-script::Environment` once we're in a place to move that to it's own crate.

.github/workflows/build.yaml
```diff
@@ -244,6 +244,33 @@ jobs:
...
+        # Some notes on this build command:
+        # - The vendored headers are used to get the necessary DCAP headers
+        # - The installed `tlibc` is used to get a compilable `time.h` for the target.
+        # - In the unlikely event that `thumbv7m-none-eabi` was installed with rustup, this would error out with
+        #   duplicate core symbols due to `-Z build-std=core`.
+        run: |
+          cargo metadata --no-deps --format-version=1 |  \
+            jq -r '.packages[].name' | \
+            grep -e types | \
+            xargs -n1 sh -c 'CFLAGS="-isystem${GITHUB_WORKSPACE}/core/build/headers -isystem/opt/intel/sgxsdk/include/tlibc" cargo build -Z build-std=core --target thumbv7m-none-eabi -p $0 || exit 255'
```
jcape, line 272: How do you feel about vendoring the tlibc headers rather than depending on `/opt/intel/sgxsdk/include/tlibc`?

Cargo.toml
jcape: Does `getrandom` use `edition = "2021"`?

jcape (commented):
LGTM, but I'd like to think about a few things:

Cargo.toml
nick-mobilecoin: No, I added 2021 to `rand`, `getrandom`, and `rand_core` and it still resulted in the same error
```shell
error[E0463]: can't find crate for `std`
 --> /home/nick/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.7/src/error_impls.rs:9:1
  |
9 | extern crate std;
  | ^^^^^^^^^^^^^^^^^ can't find crate
  |
  = note: the `thumbv7m-none-eabi` target may not support the standard library
  = help: consider building the standard library from source with `cargo build -Zbuild-std`
```

.github/workflows/build.yaml
nick-mobilecoin: vendored the tlibc headers, but still used the cflags.

I tried being smart about sending the tlibc if the SGX_SDK isn't found, but this got a bit nasty around the test enclave.
I'm fine if we want to rethink how the test enclave gets those paths, but implictly adding tlibc when SGX_SDK not found also felt like an odd inconsistent behavior.

core/build/src/lib.rs
nick-mobilecoin: I'm interpreting this that we're fine for whitespace here at this time, but we shouldn't shy away from shlex going forward.

leaving as is unless I hear otherwise.

jcape (approved):

PR #182, opened 2022-10-24
Pull our custom actions from the MCF actions repo.

jcape:
As it says on the tin, this depends on mobilecoinfoundation/actions#1 (henceforth referred to as "Action Comics No. 1) getting approved and merged.


jcape:
* main
  * **PR #182** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/182" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/182?utm_source=stack-comment).

.github/workflows/build.yaml
```diff
@@ -112,7 +112,7 @@ jobs:
           - nightly-2022-07-22
     steps:
       - uses: actions/checkout@v3
-      - uses: ./.github/actions/dcap-libs
+      - uses: mobilecoinfoundation/actions/dcap-libs@main
```
nick-mobilecoin, line 115: Are we thinking for now we pull from main and then once things settle we start following tags?

jcape: I'm a bit torn on this generally, TBH. If we want to version stuff our best bet is to put them in separate repos, e.g. `mobilecoinfoundation/action-install-dcap@v2` would be versioned differently from `mobilecoinfoundation/install-sgx-sdk@v2.17.101.1`

nick-mobilecoin: Yah, I was also wondering if the sdk version string should be an input.
I.e. we had to hold back on the sdk for one of the branches for a little bit until we had time to switch over.

jcape: Yeah, I figure we can leave them in the one repo now, and fiddle with things as necessary. I don't expect us to be changing this stuff all that often?

awygle (approved):

nick-mobilecoin (approved):

PR #184, opened 2022-10-31
Use stable for rust builds

nick-mobilecoin:
Previously a nightly version of rust were used for CI and locally, now
the stable version is used.

The nighly is still used for the no alloc CI builds as these builds require the `-Z` flag which is nightly only.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -14,15 +14,11 @@ env:
 jobs:
   rustfmt:
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        rust:
-          - nightly-2022-07-22
     steps:
       - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: ${{ matrix.rust }}
+          toolchain: stable
           profile: minimal
           components: rustfmt
           override: true
@@ -106,17 +102,13 @@ jobs:
     needs:
       - "rustfmt"
       - "markdown-lint"
-    strategy:
-      matrix:
-        rust:
-          - nightly-2022-07-22
     steps:
       - uses: actions/checkout@v3
       - uses: mobilecoinfoundation/actions/dcap-libs@main
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: ${{ matrix.rust }}
+          toolchain: stable
           components: clippy
           override: true
       - uses: r7kamura/rust-problem-matchers@v1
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* main
  * **PR #184** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/184" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/184?utm_source=stack-comment).

awygle (approved):

.github/workflows/build.yaml
```diff
@@ -106,17 +102,13 @@ jobs:
...
-        rust:
-          - nightly-2022-07-22
     steps:
       - uses: actions/checkout@v3
       - uses: mobilecoinfoundation/actions/dcap-libs@main
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: ${{ matrix.rust }}
+          toolchain: stable
```
jcape, line 111: For both `clippy` and `rustfmt`, we should actually fail on `stable`, but generate warnings for `beta`---the most likely thing to cause a CI fail on a stable upgrade is going to be `clippy` adding a lint.

Doesn't need to happen in this PR, but it should in a follow-up.

.github/workflows/build.yaml
```diff
@@ -157,7 +149,9 @@ jobs:
     strategy:
       matrix:
         rust:
-          - nightly-2022-07-22
+          - stable
+          - beta
+          - nightly
```
jcape, line 154: Nightly should be a pinned version that we standardize across all the MCF repos (i.e. whatever we're going to build enclaves with)

.github/workflows/build.yaml
```diff
@@ -208,8 +204,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          # FIXME: replace with stable ASAP
-          toolchain: nightly-2022-07-22
+          toolchain: stable
```
jcape, line 207: This should be the same "stable = fail, beta = warn" scenario, though nightly won't be required since docs.rs won't be using it.

jcape (approved):
LGTM, but we'll want a follow-up PR to create "warn-if-fail" builds on `beta` and a specific nightly that we do bless for everything in the MCF org.

.github/workflows/build.yaml
nick-mobilecoin: pinned back to 2022-07-22.
Note that the main mobile coin repo is using 2022-04-29, but I've got a hunch that will get updated prior to these crates getting brought into that repo.

.github/workflows/build.yaml
nick-mobilecoin: Created #187

PR #189, opened 2022-11-04
Remove nono job from CI

nick-mobilecoin:
Previously the CI builds utilized
[cargo nono](https://github.com/hobofan/cargo-nono). This had some
false positives when depending on the rand crate.  The `build-no-alloc`
CI job fulfills the same propose in a more robust way since it builds
the crates.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,3 +1,1 @@
 edition = "2021"
-wrap_comments = true
-imports_granularity = "Crate"
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -117,30 +117,6 @@ jobs:
           command: clippy
           args: --all --all-features --locked -- -D warnings
 
-  nono:
-    runs-on: ubuntu-20.04
-    needs:
-      - "rustfmt"
-      - "markdown-lint"
-    steps:
-      - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-      - uses: actions-rs/install@v0.1
-        with:
-          crate: cargo-nono
-          version: latest
-          use-tool-cache: true
-      - uses: r7kamura/rust-problem-matchers@v1
-      - run: |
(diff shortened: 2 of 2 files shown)


nick-mobilecoin:
* main
  * **PR #189** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/189" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/189?utm_source=stack-comment).

rustfmt.toml
```diff
@@ -1,3 +1 @@
 edition = "2021"
-wrap_comments = true
-imports_granularity = "Crate"
```
nick-mobilecoin, line 3: I noticed since the change to defaulting to stable that these options don't work locally and spam the terminal unless one adds the `+nightly` arg and
```shell
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
...
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
```

jcape: Blech. OK.

https://github.com/rust-lang/rustfmt/issues/4991

jcape (approved):

rustfmt.toml
nick-mobilecoin: thanks for creating the issues, I should have done that instead of hoping one of us would remember to add it back...

NotGyro (approved):
Looks good, this should be cleanly removed.

PR #190, opened 2022-11-04
Rename `mc_sgx_dcap_types::Quote3Error` to SgxError

nick-mobilecoin:
Previously the error coming back from the SGX SDK for quote(3) related
functionality was named `mc_sgx_dcap_types::Quote3Error`. This has been
renamed to `mc_sgx_dcap_types::SgxError` to better indicate it is an
error coming from the SGX SDK.

### DIFF
--- a/dcap/types/src/error.rs
+++ b/dcap/types/src/error.rs
@@ -16,7 +16,7 @@ use serde::{Deserialize, Serialize};
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[non_exhaustive]
 #[repr(u32)]
-pub enum Error {
+pub enum SgxError {
     /// An unexpected internal error occurred
     Unexpected = quote3_error_t::SGX_QL_ERROR_UNEXPECTED.0,
     /// One of the parameters passed to an SGX FFI method was invalid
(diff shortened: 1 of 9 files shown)


dcap/types/src/error.rs
```diff
@@ -16,7 +16,7 @@ use serde::{Deserialize, Serialize};
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[non_exhaustive]
 #[repr(u32)]
-pub enum Error {
+pub enum SgxError {
```
nick-mobilecoin, line 19: Another option is to rename the module `sgx_error.rs` and keep the enum named `Error` then rename in the public API like the `Error` -> `Quote3Error` that was happening.

This naming issue was noticed as the Quote3 type started to gain fallible behavior. It feels off that the Quote3Error doesn't correlate to a failure of the Quote3 type. The Quote3Error naming was mimicing the SGX SDK's `quote3_error_t`.

nick-mobilecoin: This has been further renamed to `QlError` leaving full comment open for context/discussion.

jcape (approved):
LGTM, but the `Error::Sgx` -> `Error::QuoteLib` change entails some corresponding doc updates (see annotations).

awygle (approved):
LGTM.

PR #194, opened 2022-11-08
Add `mc_sgx_dcap_types::QuoteHeader`

nick-mobilecoin:

PR #193, opened 2022-11-07
Update Quote3 to ensure correct version and size.

nick-mobilecoin:
Previously Quote3 was using the `From` trait and did not verify that the
source bytes were sufficient to represent a Quote3. Now Quote3 uitilizes
the `TryFrom` trait and errors if the source bytes are not large enough
or do not represent a version 3 quote.

Issue #66, opened 2022-08-15
Handle `Quote` (specifically Quote3)


jcape:
- [x] Look at pulling nom in for converting bytes to the types

nick-mobilecoin:
Discussion from 2022-08-16:
* Create a rust type that holds a `Vec<u8>`. This will be the raw quote.  
  * The need to have this raw representation is to be able to verify the signatures on parts of these raw bytes.
* Create a rust type for each of the SGX quote types, `sgx_quote_t`, `sgx_quote3_t`, and `sgx_quote4_t`.
  * Initial thought was a view, but with https://github.com/rust-lang/rust/issues/82523 we think we'll need to copy out instead of being able to pass references.
  * `sgx_quote_t` looks to be EPID, `sgx_quote3_t` looks to be DCAP, and `sgx_quote4_t` looks to be for TDX

jcape:
Thoughts as of today's sprint-planning: Quote3 is a rustic structure that contains the parsed Quote bytes, as well as the underlying bytes (as slice or vec).

nick-mobilecoin:
Implemented via #190, #196, #197, #200, #202, #203, #204, #205, #208

PR #195, opened 2022-11-08
Add `From` bytes for `QuoteHeader`

nick-mobilecoin:
Add functionality to unpack a `QuoteHeader` from a byte slice.

### DIFF
--- a/dcap/types/src/quote_header.rs
+++ b/dcap/types/src/quote_header.rs
@@ -2,6 +2,7 @@
 
 //! This module provides types related to Quote v3
 
+use core::mem;
 use mc_sgx_core_types::{new_type_accessors_impls, Algorithm, FfiError, IsvSvn};
 use mc_sgx_dcap_sys_types::sgx_quote_header_t;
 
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
* main
  * **PR #193** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/193" target="_blank">Graphite</a> 
    * **PR #194** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/194" target="_blank">Graphite</a> 
      * **PR #195** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/195" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/195?utm_source=stack-comment).

dcap/types/src/quote_header.rs
```diff
@@ -54,9 +55,88 @@ new_type_accessors_impls! {
...
+    fn from(bytes: [u8; mem::size_of::<sgx_quote_header_t>()]) -> Self {
+        // A note about the `expect()` calls.  This size is specified in the
+        // signature and there are unit tests ensuring the extraction from a
+        // byte array.  The values should not fail to extract due to size
+        // issues.
+        let mut header = Self::default();
+        header.0.version =
+            u16::from_le_bytes(bytes[..2].try_into().expect("Failed to extract `version`"));
+        header.0.att_key_type = u16::from_le_bytes(
+            bytes[2..4]
```
nick-mobilecoin, line 68: I've been thinking about magic numbers like these and in use in ReportBody's `From` implementation.

I'm not fond of making a bunch of defines for all these value when there only used here as it creates  a level of indirection and doesn't ensure correctness.
One thought I had to help address this is to come up with some kind of incrementable range:
```rust
let range = Magic::new(0, mem::sizeof::<u16>());
header.0.att_key_type = u16::from_le_bytes(bytes[range].try_into()..expect("blah");
let range = Magic::new_from_other(range, mem::sizeof::<u32>())
header.0.att_key_data_0 = u32::from_le_bytes(bytes[range].try_into()..expect("blah");
```
IDK it feels a bit clunky, but thought I would throw it out there in case someone has thoughts on:
- ways to make it more palatable
- arguments exposing the wrongness of this idea

jcape: I think for this specific object, there's no reason to worry about validating the data, since that decision can be punted until the user actually tries to create a versioned quote object from the bytes underlying this structure.

What I would suggest is flipping this problem on it's head: each `QuoteN` variant actually implements a trait that contains what this magic number should be, e.g.:

```rust
pub trait VersionedQuote {
    const VERSION: u16;
    type Err: Display;

    fn try_from_header(header: &QuoteHeader, bytes: &[u8]) -> Result<Self, Self::Error> {
    }
}
```

jcape: Oh, I see, you're talking about the byte ranges...

Is this something `nom` could solve?

nick-mobilecoin: nom may be the way to go.
That's probably a better way to think of it; a consumed stream or a cursor of some kind

nick-mobilecoin: I've created a task on #66 to pull in nom.  I think it can be done as a follow-on PR.

samdealy (approved):

nick-mobilecoin:
Since the Quote header type has changed in each quote version it was decided that exposing it outside of the quote3 type will not provide much benefit.

PR #196, opened 2022-11-14
Change `mc-sgx-core-types::ReportBody` to use TryFrom

nick-mobilecoin:
Previously `mc-sgx-core-types::ReportBody` used the from trait for a
fixed size slice. Now the `mc-sgx-core-types::ReportBody` uses the
TryFrom trait on a dynamically sized slice.

The fixed sized slice required copying from a dynamic buffer into a
temporary fixed sized buffer in order to fullfil the compile
time constraint of the fixed size.

### DIFF
--- a/core/types/src/report.rs
+++ b/core/types/src/report.rs
@@ -3,14 +3,18 @@
 
 use crate::{
     config_id::ConfigId, impl_newtype_for_bytestruct, key_request::KeyId, new_type_accessors_impls,
-    Attributes, ConfigSvn, CpuSvn, IsvSvn, Measurement, MiscellaneousSelect, MrEnclave, MrSigner,
+    Attributes, ConfigSvn, CpuSvn, FfiError, IsvSvn, Measurement, MiscellaneousSelect, MrEnclave,
+    MrSigner,
 };
-use core::mem;
 use mc_sgx_core_sys_types::{
     sgx_isvext_prod_id_t, sgx_isvfamily_id_t, sgx_mac_t, sgx_prod_id_t, sgx_report_body_t,
     sgx_report_data_t, sgx_report_t, SGX_CONFIGID_SIZE, SGX_CPUSVN_SIZE, SGX_HASH_SIZE,
-    SGX_ISVEXT_PROD_ID_SIZE, SGX_ISV_FAMILY_ID_SIZE, SGX_REPORT_DATA_SIZE,
+    SGX_ISVEXT_PROD_ID_SIZE, SGX_ISV_FAMILY_ID_SIZE, SGX_REPORT_BODY_RESERVED1_BYTES,
+    SGX_REPORT_BODY_RESERVED2_BYTES, SGX_REPORT_BODY_RESERVED3_BYTES,
+    SGX_REPORT_BODY_RESERVED4_BYTES, SGX_REPORT_DATA_SIZE,
 };
+use nom::bytes::complete::take;
+use nom::number::complete::{le_u16, le_u32, le_u64};
 
 /// MAC
 #[derive(Default, Debug, Clone, Hash, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


nick-mobilecoin:
* main
  * **PR #193** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/193" target="_blank">Graphite</a> 
    * **PR #194** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/194" target="_blank">Graphite</a> 
      * **PR #195** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/195" target="_blank">Graphite</a> 
    * **PR #196** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/196" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/196?utm_source=stack-comment).

core/types/src/report.rs
```diff
@@ -3,14 +3,18 @@
...
 use mc_sgx_core_sys_types::{
     sgx_isvext_prod_id_t, sgx_isvfamily_id_t, sgx_mac_t, sgx_prod_id_t, sgx_report_body_t,
     sgx_report_data_t, sgx_report_t, SGX_CONFIGID_SIZE, SGX_CPUSVN_SIZE, SGX_HASH_SIZE,
-    SGX_ISVEXT_PROD_ID_SIZE, SGX_ISV_FAMILY_ID_SIZE, SGX_REPORT_DATA_SIZE,
+    SGX_ISVEXT_PROD_ID_SIZE, SGX_ISV_FAMILY_ID_SIZE, SGX_REPORT_BODY_RESERVED1_BYTES,
+    SGX_REPORT_BODY_RESERVED2_BYTES, SGX_REPORT_BODY_RESERVED3_BYTES,
+    SGX_REPORT_BODY_RESERVED4_BYTES, SGX_REPORT_DATA_SIZE,
 };
+use nom::bytes::complete::take;
+use nom::number::complete::{le_u16, le_u32, le_u64};
```
nick-mobilecoin, line 17: Up for suggestions the [convention](https://doc.rust-lang.org/1.0.0/style/style/imports.html#prefer-fully-importing-types/traits-while-module-qualifying-functions.) is to module qualify functions, but for `nom` the would be `complete::x` which would be ambiguous.

core/types/src/report.rs
```diff
@@ -128,72 +132,67 @@ new_type_accessors_impls! {
...
+        let (bytes, config_svn) = le_u16(bytes)?;
+        body.0.config_svn = config_svn;
+
+        let (bytes, reserved4) = take(SGX_REPORT_BODY_RESERVED4_BYTES)(bytes)?;
+        body.0.reserved4.as_mut().copy_from_slice(reserved4);
+
+        let (bytes, isv_family_id) = take(SGX_ISV_FAMILY_ID_SIZE)(bytes)?;
+        body.0.isv_family_id.as_mut().copy_from_slice(isv_family_id);
+
+        let (_, report_data) = take(SGX_REPORT_DATA_SIZE)(bytes)?;
```
awygle, line 192: Are we OK with the input hypothetically being too long? Or should it return an error? I suppose I am OK with 'too long' in this case as this is `TryFrom<&[u8]>` and therefore doesn't take the bytes, but it's a bit odd...

nick-mobilecoin: Initially I was leaning toward *not* barking on too long.
The main reason being that the report body is often in the middle of a byte stream so this would force callers to end the slice.
However since this `TryFrom` doesn't consume from the slice, like nom, the callers still end up having to jump over the report body to the next item in the stream so it doesn't gain much.

In general I would say don't place an unnecessary constraint unless we can think of a reason, but I'm not fully sold here either.

NotGyro (approved):
Looks good

nick-mobilecoin:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/196?utm_source=stack-comment).


Job Reference: yIELMGRp3yYkMM34LUYh

PR #198, opened 2022-11-15
Remove overriding profile defaults

nick-mobilecoin:
Previously the build profiles for dev and release had some values being
overriden. Now the default profile values are used.

In particular the dev profile was setting `lto` to true, which caused
issues debugging tests. There was no previous motivation for the
overrides, they were copied from an earlier implementation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,17 +45,6 @@ exclude = [
 # Even specifying `edition = "2021"` here will not fix this
 resolver = "2"
 
-[profile.dev]
-opt-level = 0
-lto = true
-
-[profile.release]
-opt-level = 3
-rpath = false
-lto = true
-debug-assertions = false
-overflow-checks = false
-
 [workspace.metadata.release]
 shared-version = true
 dev-version-ext = "beta.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,9 @@ exclude = [
 # Even specifying `edition = "2021"` here will not fix this
 resolver = "2"
 
+[profile.release]
+lto = true
+
 [workspace.metadata.release]
 shared-version = true
 dev-version-ext = "beta.0"


Cargo.toml
```diff
@@ -45,17 +45,6 @@ exclude = [
 # Even specifying `edition = "2021"` here will not fix this
 resolver = "2"
 
-[profile.dev]
-opt-level = 0
-lto = true
```
nick-mobilecoin, line 50: FYI: @jcape having lto on was causing difficulties debugging test runs.  
This looks to go all the way back to https://github.com/mobilecoinfoundation/sgx/commit/40838d72c1f0e477f36bfb4703793ce404190bcf

nick-mobilecoin: I did not, I think for `dev` we should avoid any kind of optimizations to make it easier to debug. 
It could just be me and my history with c/c++, but I think most people would expect step over and step in to do it in debug/dev builds.  Only if someone did a release with debug symbols or similar do they expect to skip over code when degbugging.

Looking at [thin](https://clang.llvm.org/docs/ThinLTO.html) it seems that this is how much LTO happens and also provides an incremental linking capability, like was added in MSVC2019.

awygle: Yeah I'm more worried about the release profile - the default `--release` doesn't include LTO at all.

awygle (approved):

jcape (approved):
LGTM

Cargo.toml
nick-mobilecoin: Added lto back into the release profile

nick-mobilecoin:
**Graphite Merge Job**

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/198?utm_source=stack-comment).


Job Reference: yIELMGRp3yYkMM34LUYh

PR #200, opened 2022-11-17
Use nom for Quote3 parsing

nick-mobilecoin:
Also add better testing around the `AuthenticationData` and
`CertificationData`

### DIFF
--- a/dcap/types/src/quote3.rs
+++ b/dcap/types/src/quote3.rs
@@ -9,6 +9,7 @@ use core::hash::{Hash, Hasher};
 use core::mem;
 use mc_sgx_core_types::{QuoteNonce, ReportBody, ReportData};
 use mc_sgx_dcap_sys_types::{sgx_ql_ecdsa_sig_data_t, sgx_quote3_t, sgx_quote_header_t};
+use nom::number::complete::{le_u16, le_u32};
 use sha2::{Digest, Sha256};
 use static_assertions::const_assert;
 use subtle::ConstantTimeEq;
@@ -29,11 +30,19 @@ const AUTH_DATA_OFFSET: usize = QUOTE_SIZE;
 // The offset to the report body for the app. From the start of the quote.
 const REPORT_BODY_OFFSET: usize = mem::size_of::<sgx_quote_header_t>();
 
+/// The minimum size of a byte array to contain a [`AuthenticationData`]
+/// the 2 bytes for QE authentication data size
+const MIN_AUTH_DATA_SIZE: usize = 2;
+
+/// The minimum size of a byte array to contain a [`CertificationData`]
+/// The 2(type) + 4(size) for QE certification data
+const MIN_CERT_DATA_SIZE: usize = 6;
+
 /// The minimum size of a byte array to contain a [`Quote3`]
-///
-// 8 is from the 2 bytes for QE authentication data size and 2(type) + 4(size)
-// for QE certification data
-pub const MIN_QUOTE_SIZE: usize = QUOTE_SIZE + 8;
+pub const MIN_QUOTE_SIZE: usize = mem::size_of::<sgx_quote3_t>()
+    + mem::size_of::<sgx_ql_ecdsa_sig_data_t>()
+    + MIN_AUTH_DATA_SIZE
+    + MIN_CERT_DATA_SIZE;
 
 /// Errors interacting with a Quote3
 #[derive(Clone, Debug, displaydoc::Display, Eq, Hash, PartialEq, PartialOrd, Ord)]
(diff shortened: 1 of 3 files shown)


nick-mobilecoin:
* main
  * **PR #200** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/200" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/200?utm_source=stack-comment).

dcap/types/src/quote3.rs
```diff
@@ -65,20 +74,34 @@ impl Error {
...
 
+impl From<nom::Err<nom::error::Error<&[u8]>>> for Error {
+    fn from(_: nom::Err<nom::error::Error<&[u8]>>) -> Self {
+        // nom errors rely on the lifetime of the data being parsed, so we
+        // convert.
+        // nom errors should only happen if the size was wrong for decoding the
+        // quote.
+        Error::InputLength {
+            required: MIN_QUOTE_SIZE,
+            actual: 0,
```
nick-mobilecoin, line 85: Not fond of this actual size being 0 here.
Also not fond of that this relies on all nom usages to have  preguard ensuring nom shouldn't fail.
Up for suggestions here.  Could change nom usages to `except()` statements, but errors seem like a nicer behavior for customers.

dcap/types/src/quote3.rs
```diff
@@ -208,32 +233,39 @@ impl TryFrom<Vec<u8>> for Quote3<Vec<u8>> {
 /// Table 8 of
 /// <https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf>.
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
-struct AuthenticationData<'a> {
-    bytes: &'a [u8],
-    // Since this has to be read always we unpack for availability.
-    // This is the `size` field, *not* the length of `bytes`.
-    data_size: usize,
+struct AuthenticationData<T> {
```
jcape, line 236: What's the goal behind switching this to a type param?

nick-mobilecoin: Reverted back to lifetime on a slice

dcap/types/src/quote3.rs
nick-mobilecoin: Changed moving discusion to comment on new code

dcap/types/src/quote3.rs
```diff
@@ -65,20 +74,33 @@ impl Error {
...
+// ```rust
+//  let (bytes, value) = le_u16::<_, nom::error::Error<&[u8]>>(bytes)
+//      .expect("Should have been prevented by previous size check.");
+// ```
+impl From<nom::Err<nom::error::Error<&[u8]>>> for Error {
+    fn from(_: nom::Err<nom::error::Error<&[u8]>>) -> Self {
+        panic!("Nom errors should be prevented by size checks when parsing a Quote3");
+    }
+}
+
```
nick-mobilecoin, line 88: Thoughts on this?
I felt specifying the error type in each nom invocation was a bit heavy handed so left the error conversion to have it panic.

nick-mobilecoin: Also if there is some rustism I'm missing to make this cleaner, I'm all for that.

jcape: The problem is that this conversion is global, but the checks that make it OK are local. So, like, if someone later uses `Quote3Error` as their return type, and also use `nom` for something else, they can trigger a panic even though they may want a handled error.

IMO the best option here is to write a quick macro for this, e.g. `nom_expect!(tt)` that appends the `.expect("Nom errors...")`, and trust that LLVM will handle the static string properly.

nick-mobilecoin: Added wrapper functions. I didn't look too far into the macro implementation but I think it would end up being fragile by needing to break up the function name and parens to insert the error type. 
It's possible to macro the 2 wrapper functions written, but it felt like the downside of obfuscation outweighed the just do it twice.

awygle (approved):

dcap/types/src/quote3.rs
```diff
@@ -150,39 +159,40 @@ impl<T: AsRef<[u8]>> Quote3<T> {
...
         // Similar to above this shouldn't fail since we checked for `MIN_QUOTE_SIZE` above.
         let report_body =
-            ReportBody::try_from(&bytes.as_ref()[REPORT_BODY_OFFSET..]).map_err(|_| {
-                Error::InputLength {
-                    required: MIN_QUOTE_SIZE,
-                    actual: bytes_length,
-                }
+            ReportBody::try_from(&bytes[REPORT_BODY_OFFSET..]).map_err(|_| Error::InputLength {
+                required: MIN_QUOTE_SIZE,
+                actual: bytes_length,
```
jcape, line 181:
```suggestion
                actual,
```

dcap/types/src/quote3.rs
```diff
@@ -150,39 +159,40 @@ impl<T: AsRef<[u8]>> Quote3<T> {
...
     fn try_from_bytes(bytes: T) -> Result<Self> {
-        let ref_bytes = bytes.as_ref();
-        let bytes_length = ref_bytes.len();
+        let raw_bytes = bytes;
+        let bytes = raw_bytes.as_ref();
+        let bytes_length = bytes.len();
         if bytes_length < MIN_QUOTE_SIZE {
             return Err(Error::InputLength {
                 required: MIN_QUOTE_SIZE,
                 actual: bytes_length,
```
jcape, line 168:
```suggestion
        let actual = bytes.len();
        if actual < MIN_QUOTE_SIZE {
            return Err(Error::InputLength {
                required: MIN_QUOTE_SIZE,
                actual,
```

jcape (approved):
LGTM, w/ minor nits

PR #202, opened 2022-11-17
Add SignatureData

nick-mobilecoin:
Add a new type to contain the members of the Quote3 signature data,
`sgx_ql_ecdsa_sig_data_t`.

PR #203, opened 2022-11-18
Add functionality to iterate over PEM certs

nick-mobilecoin:
The `CertificationData` can now be turned into an iterator that provides
each PEM certificate in the data.


nick-mobilecoin:
* main
  * **PR #200** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/200" target="_blank">Graphite</a> 
    * **PR #202** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/202" target="_blank">Graphite</a> 
      * **PR #203** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/203" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/203?utm_source=stack-comment).

dcap/types/src/quote3.rs
```diff
@@ -367,6 +367,60 @@ impl<'a> TryFrom<&'a [u8]> for CertificationData<'a> {
     }
 }
 
+impl<'a> IntoIterator for &'a CertificationData<'a> {
```
nick-mobilecoin, line 370: I'm thinking with the addition of this, if this is the desired path, the `CertificationData` should get moved out to it's own module, but in a follow on PR that only does that.

dcap/types/src/quote3.rs
```diff
@@ -367,6 +367,60 @@ impl<'a> TryFrom<&'a [u8]> for CertificationData<'a> {
...
+        if self.pem_data.is_empty() {
+            return None;
+        }
+
+        let mut start = None;
+        let mut offset = 0;
+
+        // The data comes across the wire and should only use "\n" and thus
+        // won't be platform specific
+        let lines = self.pem_data.split_inclusive(|e| *e == b'\n');
```
nick-mobilecoin, line 401: I looked for a PEM _walking_ crate that was `no_std` but my google fu came up empty

nick-mobilecoin: Looking at `pem-rs` it seems the behavior is similar, https://github.com/jcreekmore/pem-rs/blob/master/src/parser.rs#L36
`pem-rs` will skip over another BEGIN line.  One difference is that `pem-rs` will consume mismatched END lines and then error at https://github.com/jcreekmore/pem-rs/blob/master/src/lib.rs#L163

dcap/types/src/quote3.rs
```diff
@@ -378,6 +378,67 @@ impl<'a> TryFrom<&'a [u8]> for CertificationData<'a> {
...
+
+struct PemIterator<'a> {
+    pem_data: &'a [u8],
+}
+
+/// Iterator over each PEM in a provided buffer.
+impl<'a> Iterator for PemIterator<'a> {
+    type Item = &'a [u8];
+    fn next(&mut self) -> Option<&'a [u8]> {
+        if self.pem_data.is_empty() {
```
NotGyro, line 403: Is there an expected, normal case where a `CertificationData` will hold no PEM certs at all? It seems like maybe this should be an error rather than silent failure but I don't fully understand how we're using this so it might make sense to just skip over it if it's empty here.

dcap/types/src/quote3.rs
NotGyro: The ergonomics of an iterator that returns `Option<Result<T>>` rather than `Option<T>` always feel pretty clunky to me. Maybe a `parse_multi_pem()` method on `&'a CertificationData<'a>` that returns a `Result<Vec<&'a [u8]>>` which you then iterate on would be better?
If this sounds like it would produce worse code here feel free to shoot me down on this, I am not speaking with absolute confidence here.

dcap/types/src/quote3.rs
nick-mobilecoin: I think for the _generic_ `PemIterator` it makes sense. (though I may be guilty of genericising too soon)

It may be that instead of the `IntoIterator` trait for `CertificationData` it is refactored to a fallible method of `into_pem_iter()` or similar.  
Though this may still be lacking full assurances.  At some point we will need to convert `PEM -> DER` and then verify the resultant cert chain.  I think the verification of the cert chain will be the most logical place to fail, no cert chain should be an "unverifiable" error.

dcap/types/src/quote3.rs
nick-mobilecoin: Definitely in a grey area here. The `type` crates we want to support `no_alloc`.  It doesn't mean we can't provide interfaces hidden behind an `alloc` feature. But _ideally_ it would be nice to find a common interface for the `alloc` and `no_alloc` use cases. Though how much time do we spend trying to make that happen versus get something working, is always subjective.

With that said anything using `Vec` would be isolated to the `alloc` feature only. We may be ok saying no alloc can use the `CertifictionData::data()` method added in #205. However raw access to `data` doesn't let one get at the PEMs in a meaningful way.  

When I initially started implementing this I tried to go straight to DER with https://docs.rs/pem-rfc7468/latest/pem_rfc7468/.  However in order to use https://docs.rs/pem-rfc7468/latest/pem_rfc7468/ in a `no_alloc` manner one must pass in the `&[u8]` buffer to contain the DER bytes. This works, but it feels clunky. Because I already had to parse the PEMs in order to feed them to https://docs.rs/pem-rfc7468/latest/pem_rfc7468/ I decided to only do the PEM iteration (at this time).
I still can't decide if the DER behavior should be available in this crate, or if it can be left to consumers.

NotGyro: Ah, I didn't realize this needed to be `no_alloc` - in that case I think you're implementing this in the best way available, yeah.

NotGyro (approved):
Looks good to me after clarifying (in review comments) that this needs to work in a no-std environment

PR #205, opened 2022-11-21
Move `CertificationData` to its own module

nick-mobilecoin:


dcap/types/src/certification_data.rs
```diff
@@ -0,0 +1,405 @@
+// Copyright (c) 2022 The MobileCoin Foundation
+
+//! This module provides the Certification Data type and it's logic
+
+use crate::quote3::{le_u16, le_u32};
```
nick-mobilecoin, line 5: Bringing in le_u16 and le_u32 from `qutoe3.rs` is creating a circular dependency.
I'm up for suggestions on this.  One thought is moving these to a module in `mc-sgx-util`, since `mc-sgx-core` is also using nom and it could take advantage of non fallible versions.

dcap/types/src/error.rs
```diff
@@ -8,6 +8,45 @@ use mc_sgx_util::{ResultFrom, ResultInto};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
+/// Errors interacting with a Quote3
+#[derive(Clone, Debug, Display, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[non_exhaustive]
+pub enum Quote3Error {
```
nick-mobilecoin, line 14: Renamed this on the move out as the aliasing in `lib.rs` had some interesting side effects.
If one wants to document values they have to use the internal name of `Error`. This results in documentation showing the word `Error` but linking to `Quote3Error`
```shell
warning: unresolved link to `Quote3Error::InputLength`
   --> dcap/types/src/quote3.rs:174:13
    |
174 |     /// * [`Quote3Error::InputLength`] if the length of `bytes` is not large
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no item named `Quote3Error` in scope
    |
    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
```

dcap/types/src/certification_data.rs
awygle: Yeah I'd suggest lifting these utility functions out of the `quote3.rs` module. Either that or just duplicating them here since they really don't do much.

awygle (approved):

dcap/types/src/certification_data.rs
NotGyro: I could see them being generally-useful later, it might be nice to have them in mc-sgx-util.

NotGyro (approved):
Looks good to me. Moving `le_u16` and `le_u32` to `mc-sgx-util` would make sense, but it's more of a nice-to-have than a need-to-have sort of thing and so I feel safe approving these changes as-is.

jcape:
Could you rename this from `CertificationData` to something like `PemCertificates` or `PemChain`? That's what's actually being stored in here...

nick-mobilecoin:
> Could you rename this from `CertificationData` to something like `PemCertificates` or `PemChain`? That's what's actually being stored in here...

From https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf table 9
> Data required to verify the QE Report
> Signature depending on the value of the Certification Data Type:
> 1. Byte array that contains concatenation of PPID, CPUSVN, PCESVN (LE), PCEID (LE).
> 2. Byte array that contains concatenation of PPID encrypted using RSA-2048-OAEP, CPUSVN, PCESVN (LE), PCEID (LE).
> 3. Byte array that contains concatenation of PPID encrypted using RSA-3072-OAEP, CPUSVN, PCESVN (LE), PCEID (LE).
> 4. PCK Leaf Certificate
> 5. Concatenated PCK Cert Chain (PEM formatted). PCK Leaf Cert||Intermediate CA Cert|| Root CA Cert
> 6. Intel® SGX Quote (currently not supported)
> 7. PLATFORM_MANIFEST (currently not supported)

Only type 5 is a pem chain, and maybe that's a flaw/issue with this PR.  Perhaps `CertificationData` should be an enum.  We get type 3 when one does a local attestation.  I haven't seen any of the others.

PR #224, opened 2022-12-09
Convert `CertificationData` into an enum

nick-mobilecoin:
Previously `mc-sgx-dcap-types::CertificationData` was a struct
containing opaque data. Logic was built on `CertificationData` that
assumed it would be the certificate chain variant. Now
`CertificationData` can be any of the valid variants as documented in
Table 9 of
<https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf>.

### DIFF
--- a/dcap/types/src/certification_data.rs
+++ b/dcap/types/src/certification_data.rs
@@ -17,32 +17,33 @@ type Result<T> = ::core::result::Result<T, Quote3Error>;
 /// <https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf>.
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 pub enum CertificationData<'a> {
-    /// Contains:
-    /// - Platform provisioning ID(PPID)
+    /// Contains the following data:
+    /// - Platform provisioning ID (PPID)
     /// - CPU security version number (CPUSVN)
     /// - Provisioning certification enclave security version number (PCESVN)
     /// - Provisioning certification enclave ID (PCEID)
     Ppid(&'a [u8]),
 
-    /// Contains the following encrypted with RSA 2048:
-    /// - Platform provisioning ID(PPID)
+    /// Contains the following data encrypted with RSA 2048:
+    /// - Platform provisioning ID (PPID)
     /// - CPU security version number (CPUSVN)
     /// - Provisioning certification enclave security version number (PCESVN)
     /// - Provisioning certification enclave ID (PCEID)
     PpidEncryptedRsa2048(&'a [u8]),
 
-    /// Contains the following encrypted with RSA 3072:
-    /// - Platform provisioning ID(PPID)
+    /// Contains the following data encrypted with RSA 3072:
+    /// - Platform provisioning ID (PPID)
     /// - CPU security version number (CPUSVN)
     /// - Provisioning certification enclave security version number (PCESVN)
     /// - Provisioning certification enclave ID (PCEID)
     PpidEncryptedRsa3072(&'a [u8]),
 
-    /// Contains the provisioning certification key leaf certificate
+    /// Contains the provisioning certification key (PCK) leaf certificate
     Pck(&'a [u8]),
 
-    /// Contains the provisioning certification key chain
-    PckCertChain(PckCertChain<'a>),
+    /// Contains the certificate chain for the provisioning certification key
+    /// (PCK).
+    PckCertificateChain(PckCertificateChain<'a>),
 
     /// ECDSA signature auxiliary data of an Intel SGX quote
     /// See `sgx_ql_cert_key_type_t::ECDSA_SIG_AUX_DATA`
@@ -53,7 +54,7 @@ pub enum CertificationData<'a> {
 }
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
-pub struct PckCertChain<'a> {
+pub struct PckCertificateChain<'a> {
     data: &'a [u8],
 }
 
@@ -83,7 +84,7 @@ impl<'a> TryFrom<&'a [u8]> for CertificationData<'a> {
             2 => CertificationData::PpidEncryptedRsa2048(bytes),
             3 => CertificationData::PpidEncryptedRsa3072(bytes),
             4 => CertificationData::Pck(bytes),
-            5 => CertificationData::PckCertChain(PckCertChain { data: bytes }),
+            5 => CertificationData::PckCertificateChain(PckCertificateChain { data: bytes }),
             6 => CertificationData::EcdsaSignatureAuxData(bytes),
             7 => CertificationData::PlatformManifest(bytes),
             x => return Err(Quote3Error::CertificationDataType(x)),
@@ -103,14 +104,14 @@ impl<'a> CertificationData<'a> {
             Self::PpidEncryptedRsa2048(data) => data,
             Self::PpidEncryptedRsa3072(data) => data,
             Self::Pck(data) => data,
-            Self::PckCertChain(pck_cert_chain) => pck_cert_chain.data,
+            Self::PckCertificateChain(pck_cert_chain) => pck_cert_chain.data,
             Self::EcdsaSignatureAuxData(data) => data,
             Self::PlatformManifest(data) => data,
         }
     }
 }
 
-impl<'a> IntoIterator for &'a PckCertChain<'a> {
+impl<'a> IntoIterator for &'a PckCertificateChain<'a> {
     type Item = &'a [u8];
     type IntoIter = PemIterator<'a>;
 
@@ -183,30 +184,26 @@ mod test {
     #[test]
     fn zero_type_is_error() {
         let bytes = [0u8; MIN_CERT_DATA_SIZE];
-        assert_eq!(
-            CertificationData::try_from(bytes.as_slice()),
-            Err(Quote3Error::CertificationDataType(0))
-        );
+        let error = CertificationData::try_from(bytes.as_slice()).unwrap_err();
+        assert_eq!(error, Quote3Error::CertificationDataType(0));
     }
 
     #[test]
     fn type_eight_is_error() {
         let mut bytes = [0u8; MIN_CERT_DATA_SIZE];
         bytes[0] = 8;
-        assert_eq!(
-            CertificationData::try_from(bytes.as_slice()),
-            Err(Quote3Error::CertificationDataType(8))
-        );
+        let error = CertificationData::try_from(bytes.as_slice()).unwrap_err();
+        assert_eq!(error, Quote3Error::CertificationDataType(8));
     }
 
     #[parameterized(
-    one = {1, CertificationData::Ppid(&[])},
-    two = {2, CertificationData::PpidEncryptedRsa2048(&[])},
-    three = {3, CertificationData::PpidEncryptedRsa3072(&[])},
-    four = {4, CertificationData::Pck(&[])},
-    five = {5, CertificationData::PckCertChain(PckCertChain{data: &[]})},
-    six = {6, CertificationData::EcdsaSignatureAuxData(&[])},
-    seven = {7, CertificationData::PlatformManifest(&[])},
+    byte_value_one = {1, CertificationData::Ppid(&[])},
+    byte_value_two = {2, CertificationData::PpidEncryptedRsa2048(&[])},
+    byte_value_three = {3, CertificationData::PpidEncryptedRsa3072(&[])},
+    byte_value_four = {4, CertificationData::Pck(&[])},
+    byte_value_five = {5, CertificationData::PckCertificateChain(PckCertificateChain{data: &[]})},
+    byte_value_six = {6, CertificationData::EcdsaSignatureAuxData(&[])},
+    byte_value_seven = {7, CertificationData::PlatformManifest(&[])},
     )]
     fn type_is_valid(data_type: u8, expected: CertificationData) {
         let mut bytes = [0u8; MIN_CERT_DATA_SIZE];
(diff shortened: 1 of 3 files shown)


nick-mobilecoin:
* main
  * **PR #224** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/224" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/224?utm_source=stack-comment).

dcap/types/src/certification_data.rs
```diff
@@ -16,14 +16,45 @@ type Result<T> = ::core::result::Result<T, Quote3Error>;
...
+    /// - CPU security version number (CPUSVN)
+    /// - Provisioning certification enclave security version number (PCESVN)
+    /// - Provisioning certification enclave ID (PCEID)
+    PpidEncryptedRsa3072(&'a [u8]),
+
+    /// Contains the provisioning certification key leaf certificate
+    Pck(&'a [u8]),
+
+    /// Contains the provisioning certification key chain
+    PckCertChain(PckCertChain<'a>),
```
nick-mobilecoin, line 45: The use of `Cert` here deviates from the outer type using `Certification`.  I left it short to see what peoples thoughts were on shortening the use of `Certification`in types.
I'm up for suggestions:
- Spell this one out
- Rename the outer (thinking this would be best done in a follow on PR).

dcap/types/src/certification_data.rs
```diff
@@ -59,12 +97,20 @@ impl<'a> CertificationData<'a> {
     ///
     /// `Certification Data` member from Table 9 of
     /// <https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf>.
-    pub fn data(&self) -> &[u8] {
-        self.data
+    pub fn raw_data(&self) -> &[u8] {
```
nick-mobilecoin, line 100: I left this and renamed to `raw_data` to better communicate it's just the buffer.
Not sure this is needed, but it made the testing more convenient up in quote3.rs

dcap/types/src/certification_data.rs
```diff
@@ -273,10 +352,15 @@ mod test {
...
-        let cert_data = [0u8; MIN_CERT_DATA_SIZE];
+    fn iterate_over_a_empty_pck_cert_chain() {
+        let mut cert_data = [0u8; MIN_CERT_DATA_SIZE];
+        cert_data[0] = 5;
         let certification_data = CertificationData::try_from(cert_data.as_slice()).unwrap();
-        let cert_iter = certification_data.into_iter();
+        let cert_chain = match certification_data {
+            CertificationData::PckCertChain(cert_chain) => cert_chain,
+            _ => panic!("expected a PckCertChain"),
+        };
```
nick-mobilecoin, line 362: This feels kind of wordy to me to get the expected variant, but `if let` seemed even more wordy.
Not sure if there is a more concise way to get at the underlying variant

core/build/src/lib.rs
```diff
@@ -73,7 +73,7 @@ pub fn sgx_builder() -> Builder {
         .ctypes_prefix("core::ffi")
         .allowlist_recursively(false)
         .clang_args(env_c_flags())
-        .clang_arg(&format!("-I{}", include_path));
+        .clang_arg(format!("-I{}", include_path));
```
nick-mobilecoin, line 76: This should fall out when, #223 comes in

dcap/types/src/certification_data.rs
```diff
@@ -16,14 +16,45 @@ type Result<T> = ::core::result::Result<T, Quote3Error>;
 /// Table 9 of
 /// <https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf>.
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
-pub struct CertificationData<'a> {
-    // The `Certification Data` field as described in the QuoteLibReference.
-    // The length of this *will* equal the `size` field as described in the
-    // QuoteLibReference
+pub enum CertificationData<'a> {
+    /// Contains:
+    /// - Platform provisioning ID(PPID)
```
samdealy, line 21: nit: ID (PPID)

dcap/types/src/certification_data.rs
```diff
@@ -16,14 +16,45 @@ type Result<T> = ::core::result::Result<T, Quote3Error>;
...
+    /// - CPU security version number (CPUSVN)
+    /// - Provisioning certification enclave security version number (PCESVN)
+    /// - Provisioning certification enclave ID (PCEID)
+    Ppid(&'a [u8]),
+
+    /// Contains the following encrypted with RSA 2048:
```
samdealy, line 27: nit: following data

dcap/types/src/certification_data.rs
```diff
@@ -16,14 +16,45 @@ type Result<T> = ::core::result::Result<T, Quote3Error>;
...
+    /// - Platform provisioning ID(PPID)
```
samdealy, line 28: nit: same as above.

dcap/types/src/certification_data.rs
```diff
@@ -16,14 +16,45 @@ type Result<T> = ::core::result::Result<T, Quote3Error>;
...
+    /// - CPU security version number (CPUSVN)
+    /// - Provisioning certification enclave security version number (PCESVN)
+    /// - Provisioning certification enclave ID (PCEID)
+    PpidEncryptedRsa2048(&'a [u8]),
+
+    /// Contains the following encrypted with RSA 3072:
+    /// - Platform provisioning ID(PPID)
```
samdealy, line 35: nit: same as above.

dcap/types/src/certification_data.rs
samdealy: +1 to spelling it out in order to maintain consistency.

dcap/types/src/certification_data.rs
```diff
@@ -16,14 +16,45 @@ type Result<T> = ::core::result::Result<T, Quote3Error>;
...
+    Pck(&'a [u8]),
+
+    /// Contains the provisioning certification key chain
+    PckCertChain(PckCertChain<'a>),
+
+    /// ECDSA signature auxiliary data of an Intel SGX quote
+    /// See `sgx_ql_cert_key_type_t::ECDSA_SIG_AUX_DATA`
+    EcdsaSignatureAuxData(&'a [u8]),
+
+    /// Platform manifest
```
samdealy, line 51: Can you add more detail to what a PlatformManifest is?

dcap/types/src/certification_data.rs
```diff
@@ -132,13 +178,41 @@ mod test {
...
 
     #[test]
-    fn zero_certification_data() {
+    fn zero_type_is_error() {
         let bytes = [0u8; MIN_CERT_DATA_SIZE];
+        assert_eq!(
+            CertificationData::try_from(bytes.as_slice()),
+            Err(Quote3Error::CertificationDataType(0))
+        );
+    }
```
samdealy, line 190: The usage of `Err` adds an extra layer of nesting that seems unnecessary to me. Consider changing it to the following:
```suggestion
        let error = CertificationData::try_from(bytes.as_slice()).unwrap_err();
        assert_eq!(error, Quote3Error::CertificationDataType(0));       
     }     
```

dcap/types/src/certification_data.rs
```diff
@@ -132,13 +178,41 @@ mod test {
...
+
+    #[test]
+    fn type_eight_is_error() {
+        let mut bytes = [0u8; MIN_CERT_DATA_SIZE];
+        bytes[0] = 8;
+        assert_eq!(
+            CertificationData::try_from(bytes.as_slice()),
+            Err(Quote3Error::CertificationDataType(8))
+        );
```
samdealy, line 199: Same as above.

dcap/types/src/certification_data.rs
```diff
@@ -132,13 +178,41 @@ mod test {
...
+    }
+
+    #[parameterized(
+    one = {1, CertificationData::Ppid(&[])},
+    two = {2, CertificationData::PpidEncryptedRsa2048(&[])},
+    three = {3, CertificationData::PpidEncryptedRsa3072(&[])},
+    four = {4, CertificationData::Pck(&[])},
+    five = {5, CertificationData::PckCertChain(PckCertChain{data: &[]})},
+    six = {6, CertificationData::EcdsaSignatureAuxData(&[])},
+    seven = {7, CertificationData::PlatformManifest(&[])},
```
samdealy, line 209: Can you make these test names more descriptive? I can't easily tell what the numbers correspond to.

dcap/types/src/certification_data.rs
samdealy: I've come across this, and I think that this is the best way to do it.

samdealy (approved):

dcap/types/src/certification_data.rs
eranrund: I slightly better way is coming... https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#let-else-statements

dcap/types/src/certification_data.rs
nick-mobilecoin: Unfortunatly all https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf provides is
> PLATFORM_MANIFEST (currently not supported)

While it's currently not supported the fact that it's known seems like we should at least handle it even if we don't know what it is or what if anything to do with it.

dcap/types/src/certification_data.rs
nick-mobilecoin: Spelled it out, it's actually `Certificate` in this instance so spelling out might be better to disambiguate

core/build/src/lib.rs
nick-mobilecoin: rebased should be gone now

dcap/types/src/certification_data.rs
nick-mobilecoin: added "data"

dcap/types/src/certification_data.rs
nick-mobilecoin: fixed spacing and added data

dcap/types/src/certification_data.rs
nick-mobilecoin: fixed spacing

dcap/types/src/certification_data.rs
nick-mobilecoin: separated the result -> Err logic as suggested

dcap/types/src/certification_data.rs
nick-mobilecoin: updated to `byte_value_one` I'm up for suggestions on better names.  Trying to be concise and descriptive. Being descriptive it feels like "byte value of the certification data type on the wire"

dcap/types/src/certification_data.rs
nick-mobilecoin: used multi-line logic to separate out the `Err()`

dcap/types/src/certification_data.rs
jcape: This repo is built against `stable`, `beta` (stable-next, without `feature()`), and a particular nightly, so a slightly better way is technically already here :-)

dcap/types/src/certification_data.rs
```diff
@@ -16,14 +16,46 @@ type Result<T> = ::core::result::Result<T, Quote3Error>;
...
+
+    /// ECDSA signature auxiliary data of an Intel SGX quote
+    /// See `sgx_ql_cert_key_type_t::ECDSA_SIG_AUX_DATA`
+    EcdsaSignatureAuxData(&'a [u8]),
+
+    /// Platform manifest
+    PlatformManifest(&'a [u8]),
+}
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
```
jcape, line 56: This should have docs (and the crate should have `warn(missing_docs)` if it doesn't already). This can happen in a follow-up PR.

dcap/types/src/certification_data.rs
nick-mobilecoin: `nightly-2022-07-22` 
```shell
error[E0658]: `let...else` statements are unstable
   --> dcap/types/src/certification_data.rs:356:9
    |
356 | /         let CertificationData::PckCertificateChain(cert_chain) = certification_data else {
357 | |             panic!("expected a PckCertChain");
358 | |         };
    | |__________^
    |
    = note: see issue #87335 <https://github.com/rust-lang/rust/issues/87335> for more information
    = help: add `#![feature(let_else)]` to the crate attributes to enable
```
:*(

dcap/types/src/certification_data.rs
```diff
@@ -44,13 +76,20 @@ impl<'a> TryFrom<&'a [u8]> for CertificationData<'a> {
...
         }
+        let bytes = &bytes[..data_size];
+        let data = match data_type {
+            1 => CertificationData::Ppid(bytes),
+            2 => CertificationData::PpidEncryptedRsa2048(bytes),
+            3 => CertificationData::PpidEncryptedRsa3072(bytes),
+            4 => CertificationData::Pck(bytes),
+            5 => CertificationData::PckCertificateChain(PckCertificateChain { data: bytes }),
+            6 => CertificationData::EcdsaSignatureAuxData(bytes),
+            7 => CertificationData::PlatformManifest(bytes),
```
jcape, line 89: I'd like to see the discriminant here as a constant, which may tend to lead to a slightly different design:

```rust
pub struct Ppid<'a>(&'a [u8]);

impl CertificationDataKind for Ppid {
    const KIND: u16 = 1;
}

pub enum CertificationData<'a> {
    Ppid(Ppid<'a>)
}
```

jcape (approved):
LGTM, though there is some value to introducing a newtype struct around the `&[u8]` slices in the enum.

dcap/types/src/certification_data.rs
jcape: Just update that nightly, we're only using it here (mobilecoin.git is still on 2022-04-28 or similar).

dcap/types/src/certification_data.rs
nick-mobilecoin: Updated all variants to have unique types that wrap `&'a [u8]` and provide a `KIND` constant.

nick-mobilecoin:
new beta clippy warnings are addressed in #228

awygle (approved):

awygle (approved):

PR #245, opened 2023-01-04
Add `mc-sgx-sdk-tools` crate

nick-mobilecoin:

PR #251, opened 2023-01-06
Add rust wrapper for edger8r tool

nick-mobilecoin:
Initial add of the edger8r tool for use in build.rs files

### Future Work
- Add in the search_paths
- Add build() operation to compile generated C files


nick-mobilecoin:
* main
  * **PR #245** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/245" target="_blank">Graphite</a> 
    * **PR #251** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/251" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/251?utm_source=stack-comment).

sdk-tools/src/edger8r.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub fn out_dir(&mut self, out_dir: impl AsRef<Path>) -> &mut Self {
+        self.out_dir = Some(out_dir.as_ref().to_owned());
+        self
+    }
+
+    /// Generate C files using the edger8r tool
+    ///
+    /// # Returns:
+    /// The paths to the generated files.
+    pub fn generate(&self) -> Result<Edger8rFiles, Error> {
```
nick-mobilecoin, line 126: in the [MobileCoin repo](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/util/build/sgx/src/edger8r.rs#L162) this returned `Result<&Self, Error>` so that one could do
`edger8r.generate().build()`. To minimize the size of this initial PR I only brought in the `generate()` and it felt awkward returning self from it.  I'm not sure if this, `Edger8r` struct should have the logic to compile the files that seems like general C code compilation and not specific to edger8r generated code.

sdk-tools/src/edger8r.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        let output = command.output()?;
+
+        if output.status.success() {
+            Ok(self.output_file_names(out_dir))
+        } else {
+            Err(Error::Generate(
+                format!("{command:?}"),
+                String::from_utf8(output.stdout)?,
+                String::from_utf8(output.stderr)?,
```
nick-mobilecoin, line 162: I'm thinking that perhaps these should be upgraded to `from_utf8_lossy`.  I think it unlikely that the edger8r will provide invalid utf8, but if it does it seems like we would want to provide more information to users than `Error::Utf8Error`

sdk-tools/src/edger8r.rs
jcape: I'm fine to move `build()` out, but then this builder should be changed to consuming (i.e. `foo(self) -> Self` instead of `foo(&self) -> &Self`).

sdk-tools/src/edger8r.rs
jcape: Yep, agreed.

sdk-tools/src/lib.rs
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2023 The MobileCoin Foundation
+// Copyright (c) 2022-2023 The MobileCoin Foundation
 
 #![doc = include_str!("../README.md")]
 #![deny(missing_docs, missing_debug_implementations, unsafe_code)]
+pub mod edger8r;
```
jcape, line 5:
```suggestion
mod edger8r;

pub use crate::edger8r::Edger8r;
```

jcape (changes requested):
Changes look good, but if `generate()` is going to be the terminus, we should just change this to a consuming builder.

sdk-tools/src/lib.rs
nick-mobilecoin: Done

sdk-tools/src/edger8r.rs
nick-mobilecoin: done

sdk-tools/src/edger8r.rs
nick-mobilecoin: updated `generate()` to be consuming now

sdk-tools/src/edger8r.rs
```diff
@@ -0,0 +1,366 @@
...
+};
+
+/// Errors which can occur when working with the edger8r tool.
+#[derive(Debug, Display)]
+pub enum Error {
+    /// There was an error running the command: {0}
+    Io(IoError),
+    /**
+     * There was an error generating the code,
+     * command:\n{0}\nstdout:\n{0}\n\nstderr:\n{1}
```
jcape, line 33:
```suggestion
     * command:\n{0}\nstdout:\n{1}\n\nstderr:\n{2}
```

jcape (approved):
LGTM save an incorrect format string.

jcape (changes requested):
Sorry, I wasn't clear, every builder method should take `self` and return `Self`. Otherwise you need to do this:

```rust
let foo = make_foo();
foo.bar(barval).baz(bazval);
foo.generate();
```

instead of :

```rust
build_foo().bar(barval).baz(bazval).generate();
```

sdk-tools/src/edger8r.rs
nick-mobilecoin: done

sdk-tools/src/edger8r.rs
```diff
@@ -0,0 +1,368 @@
...
+#![deny(missing_docs, missing_debug_implementations, unsafe_code)]
+
+//! Edger8r Tool Wrapper
+//!
+//! The Edger8r tool ships as part of the Intel® Software Guard Extensions SDK.
+//! It generates edge routines by reading a user-provided Enclave Description
+//! Language (EDL) file. These edge routines define the interface between the
+//! untrusted application and the enclave.
+//!
+//! See https://download.01.org/intel-sgx/sgx-linux/2.18/docs/Intel_SGX_Developer_Reference_Linux_2.18_Open_Source.pdf#%5B%7B%22num%22%3A84%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C406.5%2C0%5D
```
jcape, line 13: I'm a little surprised this wasn't annotated by the `cargo doc` CI, actually:
```suggestion
//! See [Edger8r Tool Documentation](https://download.01.org/intel-sgx/sgx-linux/2.18/docs/Intel_SGX_Developer_Reference_Linux_2.18_Open_Source.pdf#%5B%7B%22num%22%3A84%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C406.5%2C0%5D)
```

jcape (approved):

sdk-tools/src/edger8r.rs
nick-mobilecoin: Done, also moved the documentation to the `Edger8r` struct since the module itself isn't public and thus the docs weren't visible.

sdk-tools/src/edger8r.rs
```diff
@@ -0,0 +1,368 @@
...
+    /// [`OutputKind::Trusted`] is used.
+    pub untrusted: Vec<PathBuf>,
+}
+
+impl Edger8r {
+    /// Create a new edger8r builder.
+    ///
+    /// # Arguments
+    /// * `edl_file` - The edl file to generate C files from
+    pub fn new(edl_file: impl AsRef<Path>) -> Self {
```
varsha888, line 85: Why is this not returning a result with and error field like in https://github.com/mobilecoinfoundation/mobilecoin/blob/master/util/build/sgx/src/edger8r.rs#L101 ?

nick-mobilecoin: The main MobileCoin repo was using a result because of this line
```rust
let edger8r_path = get_binary(env.target_arch(), "sgx_edger8r")?;
```
This implementation utilizes [`sgx_bin_x64_dir()`](https://docs.rs/mc-sgx-core-build/0.4.0/mc_sgx_core_build/fn.sgx_bin_x64_dir.html) which doesn't fail. It returns either the `SGX_SDK` env variable or `opt/intel/sgxsdk` for the path root.
The `Command` will fail in `generate()` with "No such file or directory" if the sgx_edger8r path doesn't actually exist, https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d32be91dce2143444fa1d469ac5a4b9c

varsha888 (approved):

PR #252, opened 2023-01-06
Remove Cargo.lock

nick-mobilecoin:
Per https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
libraries shouldn't have committed lock files


nick-mobilecoin:
* main
  * **PR #252** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/252" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/252?utm_source=stack-comment).

nick-mobilecoin:
@jcape in #251 I noticed you were marked as the code owner, partly because I forgot to add the `sgx-tools` dir to the CODEOWNERS file, but also because I think the lock is getting captured in the default glob.

per https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners, we could add an entry in the `CODEOWNERS` file with no owners
```
Cargo.lock
```
However from previous discussions we said that lock files should only be present for binaries, which this repo currently doesn't have.

We may wish to leave the lock file with anticipation of binaries for sdk-tools, but that feels like YAGNI and we may change our minds.

jcape (commented):
`sgx-tools` is there mainly to provide `Command` implementations for the the signing/code-gen utilities so we can call them idiomatically from rust build scripts, so we're not going to be publishing any binaries.

But I'm fine to have make `Cargo.lock` everybody's responsibility.

jcape (approved):

varsha888 (approved):

nick-mobilecoin:
@varsha888 I bumped you for another review since this conflicted with changes that made it into `main` and this repo has "New changes require approval from someone other than author" turned on.

varsha888 (approved):

PR #253, opened 2023-01-09
Remove use of `unrwap()` in tests

nick-mobilecoin:


samdealy (approved):
LGTM. PR title has a typo in `wnrwap`.

sdk-tools/src/edger8r.rs
```diff
@@ -215,7 +215,17 @@ impl Edger8r {
 mod tests {
     use super::*;
     use std::fs;
-    use tempfile::tempdir;
+
+    /// Create a temporary directory.
```
varsha888, line 219: Why do we have to create a temp dir?

varsha888 (approved):

sdk-tools/src/edger8r.rs
nick-mobilecoin: In order to test the edger8r behavior we need files on disk both as inputs and outputs.
Using a tempdir() per test function isolates the files on disk from each test ensuring that they don't stomp on each other.
We could provide input files in a testing directory in the repo, but we would still need to isolate the output files from each other.
Another benefit of tempdir() is that it isolates each test run, so during development if you run `test_a()` that outputs `file.1` and then later you update the code or tests you know that you aren't looking at a stale `file.1` from a previous run.

PR #255, opened 2023-01-12
Expose synchronization constants and defaults

nick-mobilecoin:
Expose constants and defaults in `mc-sgx-tstdc` for mutex, rwlock, and
condition variables

### DIFF
--- a/core/build/src/lib.rs
+++ b/core/build/src/lib.rs
@@ -43,6 +43,10 @@ pub fn normalize_item_name(name: &str) -> Option<String> {
         name.push_str("_t");
     }
 
+    // Some types come through as `_sgx_thread_mutex_attr_t` which map to a real
+    // type of `sgx_thread_mutexattr_t`
+    let mut name = name.replace("_attr_", "attr_");
+
     if STRIP_UNDERSCORE_PREFIX
         .iter()
         .any(|prefix| name.starts_with(prefix))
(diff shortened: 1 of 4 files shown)


nick-mobilecoin:
* main
  * **PR #255** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/255" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/255?utm_source=stack-comment).

awygle (approved):

core/build/src/lib.rs
```diff
@@ -43,6 +43,10 @@ pub fn normalize_item_name(name: &str) -> Option<String> {
         name.push_str("_t");
     }
 
+    // Some types come through as `_sgx_thread_mutex_attr_t` which map to a real
+    // type of `sgx_thread_mutexattr_t`
+    let mut name = name.replace("_attr_", "attr_");
```
NotGyro, line 48: I was a bit nervous about false-positives here but to be fair if there was any sort of match to this pattern which shouldn't get matched then the code would just fail to compile rather than blowing up at runtime so it's probably fine.

NotGyro (approved):
This looks good to me. I was a little bit concerned about the string substitution but on further thought, if that matched anything it shouldn't that would just produce a compile error rather than blowing up at runtime, so it's fine.

core/build/src/lib.rs
nick-mobilecoin: yah, it's an interesting balance of trying to be programatic and concise vs calling out every type name.

jcape (approved):

Issue #135, opened 2022-09-16
Support testing from within an enclave


nick-mobilecoin:
There currently isn't a good way to unit test behavior that directly talks to the SGX SDK.  While it's good practice to separate logic and dependencies, it would still be good to unit test functionality at the SGX SDK boundary.

This limitation stems from the current rust test infrastructure.  The test infrastructure is invoked by the host system.  
One thought is to find a way to pass in test parameters through the host into the enclave and execute that way.
```mermaid
graph LR
    A[untrusted side] --> |ECALL| B[trusted entry point]
    D[Test runner] --> |ECALL|E[test entry point]
    subgraph Inside Enclave
    B --> C[Code to test]
    E --> F[enclave::test]
    F --> C
    end  
```

Some references:
* wasm-bindgen-test, https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/index.html#testing-on-wasm32-unknown-unknown-with-wasm-bindgen-test
* https://ferrous-systems.com/blog/test-driver-crate/#target-testing-with-defmt-test

PR #256, opened 2023-01-13
Add the `mc-sgx-tstdc` crate

nick-mobilecoin:
Add `Mutex` primitive

### DIFF
--- a/tstdc/README.md
+++ b/tstdc/README.md
@@ -1,0 +1,21 @@
+# MobileCoin: Rust wrappers around SGX synchronization primitives
+
+[![Project Chat][chat-image]][chat-link]<!--
+-->![License][license-image]<!--
+-->![Target][target-image]<!--
+-->[![Crates.io][crate-image]][crate-link]<!--
+-->[![Docs Status][docs-image]][docs-link]<!--
+-->[![Dependency Status][deps-image]][deps-link]
+
+Rust wrappers around SGX synchronization primitives
+
+[chat-image]: https://img.shields.io/discord/844353360348971068?style=flat-square
+[chat-link]: https://mobilecoin.chat
+[license-image]: https://img.shields.io/crates/l/mc-sgx-tstdc?style=flat-square
+[target-image]: https://img.shields.io/badge/target-sgx-red?style=flat-square
+[crate-image]: https://img.shields.io/crates/v/mc-sgx-tstdc.svg?style=flat-square
+[crate-link]: https://crates.io/crates/mc-sgx-tstdc
+[docs-image]: https://img.shields.io/docsrs/mc-sgx-tstdc?style=flat-square
+[docs-link]: https://docs.rs/crate/mc-sgx-tstdc
+[deps-image]: https://deps.rs/crate/mc-sgx-tstdc/0.1.0/status.svg?style=flat-square
+[deps-link]: https://deps.rs/crate/mc-sgx-tstdc/0.1.0
--- a/tstdc/src/mutex.rs
+++ b/tstdc/src/mutex.rs
@@ -1,6 +1,6 @@
 // Copyright (c) 2023 The MobileCoin Foundation
 
-//! Mutex functionality for an SGX enclave
+//! Mutex functionality for use inside of an SGX enclave
 
 use core::cell::UnsafeCell;
 use mc_sgx_tstdc_sys::{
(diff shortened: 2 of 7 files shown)


nick-mobilecoin:
* main
  * **PR #255** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/255" target="_blank">Graphite</a> 
    * **PR #256** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/256" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/256?utm_source=stack-comment).

tstdc/README.md
```diff
@@ -0,0 +1,21 @@
+# MobileCoin: Rust wrappers around SGX synchronization primitives
+
+[![Project Chat][chat-image]][chat-link]<!--
+-->![License][license-image]<!--
+-->![Target][target-image]<!--
+-->[![Crates.io][crate-image]][crate-link]<!--
+-->[![Docs Status][docs-image]][docs-link]<!--
+-->[![Dependency Status][deps-image]][deps-link]
+
+Rust wrappers around SGX synchronization primitives
```
nick-mobilecoin, line 10: Up for suggestions on this description, kind of struggled here as the library name (and thus the crate name) feels a bit odd

tstdc/src/mutex.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// # Errors
+    /// - [`Error::Busy`] if another thread has the lock, or higher precedence
+    ///   in obtaining the lock.
+    /// - [`Error::Invalid`] if self is invalid or trying to lock self when
+    ///   already holding a lock on self.
+    pub fn try_lock(&mut self) -> Result<()> {
+        let result = unsafe { sgx_thread_mutex_trylock(&mut self.0) };
+        match result {
+            0 => Ok(()),
+            libc::EBUSY => Err(Error::Busy),
```
nick-mobilecoin, line 61: Up for suggestions, not fond of an _Error_ for an ok/allowed use case

tstdc/src/mutex.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// A mutex inside of an SGX enclave
+///
+/// Use `Mutex::default()` to create an instance.
+///
+/// NB: per the documentation of
+/// [`sgx_thread_mutex_lock()`](https://download.01.org/intel-sgx/sgx-linux/2.18/docs/Intel_SGX_Developer_Reference_Linux_2.18_Open_Source.pdf#%5B%7B%22num%22%3A303%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C341.25%2C0%5D)
+/// a mutex should not be locked across root ECALLs.
+#[derive(Debug, Default, Eq, PartialEq)]
+pub struct Mutex(sgx_thread_mutex_t);
```
nick-mobilecoin, line 30: Not fond of no tests for this, but it requires running inside of an enclave, #135

tstdc/README.md
samdealy: LGTM

tstdc/src/mutex.rs
```diff
@@ -0,0 +1,100 @@
...
+
+/// A mutex inside of an SGX enclave
+///
+/// NB: per the documentation of
+/// [`sgx_thread_mutex_lock()`](https://download.01.org/intel-sgx/sgx-linux/2.18/docs/Intel_SGX_Developer_Reference_Linux_2.18_Open_Source.pdf#%5B%7B%22num%22%3A303%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C341.25%2C0%5D)
+/// a mutex should not be locked across root ECALLs.
+#[derive(Debug, Default)]
+pub struct Mutex(UnsafeCell<sgx_thread_mutex_t>);
+
+// SAFETY: The `sgx_thread_mutex_*` C functions utilize a spinlock to prevent
```
samdealy, line 31: Should this `SAFETY` block go above the struct definition on 29? I thought that's where they normally go but might be wrong.

tstdc/src/mutex.rs
```diff
@@ -0,0 +1,100 @@
...
+unsafe impl Send for Mutex {}
+unsafe impl Sync for Mutex {}
+
+impl Mutex {
+    /// Create a new non recursive mutex
+    pub const fn new() -> Self {
+        Self(UnsafeCell::new(SGX_THREAD_NONRECURSIVE_MUTEX_INITIALIZER))
+    }
+
+    /// Lock self
```
samdealy, line 42: Can you replace `self` with something more descriptive? Is it locking the underlying sgx mutex or this wrapper struct etc?

tstdc/src/mutex.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// self when already holding a lock on self.
+    pub fn lock(&self) -> Result<()> {
+        let result = unsafe { sgx_thread_mutex_lock(self.0.get()) };
+        match result {
+            0 => Ok(()),
+            _ => Err(Error::Invalid),
+        }
+    }
+
+    /// Try to lock self
```
samdealy, line 57: Same as above.

tstdc/src/mutex.rs
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2023 The MobileCoin Foundation
+
+//! Mutex functionality for an SGX enclave
```
samdealy, line 3: For my own understanding, is this `Mutex` meant to be used by multiple SGX processor threads? Or is used by multiple non-SGX threads accessing enclave code? (I may be asking the wrong question, but I'm curious as to how threading works in SGX and how this struct will be used.)

samdealy (approved):

tstdc/src/mutex.rs
nick-mobilecoin: Moved block to before struct.
I think in this case when there are multiple unsafe blocks it makes more sense.
The alternative is to document each unsafe block but they all seem to be for the same reason.

tstdc/src/mutex.rs
nick-mobilecoin: I changed to `Lock this [`Mutex`] instance` but I don't think that really addresses the concern you brought up here.
I think that the new documentation on the struct helps clarify the underlying concern and this can be left more concise.

if not let me know.

tstdc/src/mutex.rs
nick-mobilecoin: see reply above

tstdc/src/mutex.rs
nick-mobilecoin: Thanks for keeping me honest here.

I added more detail to the Mutex struct as that shows up in the generated documentation but the module documentation doesn't.
I did update to `Mutex functionality for use inside of an SGX enclave` here to try to clarify, but I think the `Mutex` docs will contain most of the docs.

Also trying to think of where to better document `root ECALL`.

tstdc/src/mutex.rs
NotGyro: It feels like the best of a bunch of bad options since Result<Option<()>> would be incredibly clunky. I don't mind it.

NotGyro (approved):
Looks good to me.

jcape (approved):

tstdc/src/mutex.rs
jcape: IMO, you probably want `Result<bool, Error>`

nick-mobilecoin: I think that makes too much sense.

Yah, it should be `Result<bool, Error>`, probably mentally latched onto mirroring the c error codes

nick-mobilecoin: Updated to a `Result<bool>` in https://github.com/mobilecoinfoundation/sgx/pull/256/commits/001eb82df774afd10e228d1ca8d694d3eb9db203

jcape (approved):

nick-mobilecoin:
⏳ This pull request is set to merge as part of a Graphite merge job
Stack job ID: oVCQ2kHpI2qOxWNUHTpX.
See details on [graphite.dev](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/256)

PR #261, opened 2023-01-25
Add `sign.rs` to `sdk-tools`

varsha888:
- Moved sign.rs file from [here](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/util/build/sgx/src/sign.rs) to sdk-tools
- Added unit tests

### DIFF
--- a/sdk-tools/src/sign.rs
+++ b/sdk-tools/src/sign.rs
@@ -1,0 +1,284 @@
+// Copyright (c) 2023 The MobileCoin Foundation
+
+//! Builder wrapper around SgxSign.
+
+use std::{
+    path::{Path, PathBuf},
+    process::Command,
+};
+
+/// Wrapper for the enclave signing tool (sgx_sign).
+///
+/// The enclave signing tool ships as part of the Intel® Software Guard
+/// Extensions SDK. Signing an enclave is a process that involves producing a
+/// signature structure that contains enclave properties such as the enclave
+/// measurement. Once an enclave is signed in such structure, the modifications
+/// to the enclave file (such as code, data, signature, and so on) can be
+/// detected. The signing tool also evaluates the enclave image for potential
+/// errors and warns you about potential security hazards.
+///
+/// See [Enclave Signing Tool Documentation](https://download.01.org/intel-sgx/sgx-linux/2.18/docs/Intel_SGX_Developer_Reference_Linux_2.18_Open_Source.pdf#%5B%7B%22num%22%3A85%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C94.5%2C206.25%2C0%5D)
+/// for more details.
+
+#[derive(Clone, Debug)]
+pub struct SgxSign {
+    /// The path to the sgx_sign executable.
+    sgx_sign_path: PathBuf,
+    /// Whether to ignore the presence of relocations in the enclave shared
+    /// object.
+    ignore_relocation_error: bool,
+    /// Whether to ignore .init sections in the enclave.
+    ignore_init_section_error: bool,
+    /// Whether to re-sign a previously signed enclave (default: false)
+    resign: bool,
+}
+
+impl Default for SgxSign {
+    /// Create a new SGX signing utility from the current environment.
+    fn default() -> Self {
+        let bin_dir = mc_sgx_core_build::sgx_bin_x64_dir();
+
+        SgxSign::from(bin_dir.join("sgx_sign"))
+    }
+}
+
+impl SgxSign {
+    /// Relocations are generally forbidden in the enclave shared object, this
+    /// tells the `sgx_sign` utility to ignore those errors.
+    #[must_use]
+    pub fn allow_relocations(mut self, allow: bool) -> Self {
+        self.ignore_relocation_error = allow;
+        self
+    }
+
+    /// Whether or not to allow .init sections in the enclave.
+    #[must_use]
+    pub fn allow_init_sections(mut self, allow: bool) -> Self {
+        self.ignore_init_section_error = allow;
+        self
+    }
+
+    /// Whether to re-sign a previously signed enclave (default: false)
+    #[must_use]
+    pub fn allow_resign(mut self, allow: bool) -> Self {
+        self.resign = allow;
+        self
+    }
+
+    /// Generate the command to sign the given enclave object with the given
+    /// private key and write the resulting enclave to the given path. Note
+    /// that online signatures are inherently insecure.
+    pub fn sign(
+        &mut self,
+        unsigned_enclave: impl AsRef<Path>,
+        config_file: impl AsRef<Path>,
+        private_key: impl AsRef<Path>,
+        output_enclave: impl AsRef<Path>,
+    ) -> Command {
+        let mut cmd = Command::new(&self.sgx_sign_path);
+        cmd.arg("sign")
+            .arg("-enclave")
+            .arg(unsigned_enclave.as_ref())
+            .arg("-config")
+            .arg(config_file.as_ref())
+            .arg("-key")
+            .arg(private_key.as_ref())
+            .arg("-out")
+            .arg(output_enclave.as_ref());
+
+        if self.ignore_relocation_error {
+            cmd.arg("-ignore-rel-error");
+        }
+
+        if self.ignore_init_section_error {
+            cmd.arg("-ignore-init-sec-error");
+        }
+
+        if self.resign {
+            cmd.arg("-resign");
+        }
+
+        cmd
+    }
+
+    /// Generate the command to create the data required for offline signing,
+    /// and write it to the given output data path.
+    pub fn gendata(
+        &mut self,
+        unsigned_enclave: impl AsRef<Path>,
+        config_file: impl AsRef<Path>,
+        output_datafile: impl AsRef<Path>,
+    ) -> Command {
+        let mut cmd = Command::new(&self.sgx_sign_path);
+        cmd.arg("gendata")
+            .arg("-enclave")
+            .arg(unsigned_enclave.as_ref())
+            .arg("-config")
+            .arg(config_file.as_ref())
+            .arg("-out")
+            .arg(output_datafile.as_ref());
+
+        if self.ignore_relocation_error {
+            cmd.arg("-ignore-rel-error");
+        }
+
+        if self.ignore_init_section_error {
+            cmd.arg("-ignore-init-sec-error");
+        }
+
+        if self.resign {
+            cmd.arg("-resign");
+        }
+
+        cmd
+    }
+
+    /// Combine an unsigned enclave and signature into the output enclave, after
+    /// checking the signature.
+    pub fn catsig(
+        &mut self,
+        unsigned_enclave: impl AsRef<Path>,
+        config_file: impl AsRef<Path>,
+        public_key_pem: impl AsRef<Path>,
+        enclave_signing_material: impl AsRef<Path>,
+        signature: impl AsRef<Path>,
+        output_enclave: impl AsRef<Path>,
+    ) -> Command {
+        let mut cmd = Command::new(&self.sgx_sign_path);
+        cmd.arg("catsig")
+            .arg("-enclave")
+            .arg(unsigned_enclave.as_ref())
+            .arg("-config")
+            .arg(config_file.as_ref())
+            .arg("-key")
+            .arg(public_key_pem.as_ref())
+            .arg("-unsigned")
+            .arg(enclave_signing_material.as_ref())
+            .arg("-sig")
+            .arg(signature.as_ref())
+            .arg("-out")
+            .arg(output_enclave.as_ref());
+
+        if self.ignore_relocation_error {
+            cmd.arg("-ignore-rel-error");
+        }
+
+        if self.ignore_init_section_error {
+            cmd.arg("-ignore-init-sec-error");
+        }
+
+        cmd
+    }
+
+    /// Examine a signed enclave file and dump the data
+    pub fn dump(
+        &mut self,
+        signed_enclave: impl AsRef<Path>,
+        css_file_path: impl AsRef<Path>,
+        dump_file_path: impl AsRef<Path>,
+    ) -> Command {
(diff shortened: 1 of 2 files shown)


sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,275 @@
...
+};
+
+/// This structure wraps the execution of sgx_sign commands.
+#[derive(Default, Clone, Debug)]
+pub struct SgxSign {
+    /// The path to the sgx_sign executable.
+    sgx_sign_path: PathBuf,
+    /// Whether to ignore the presence of relocations in the enclave shared
+    /// object.
+    ignore_rel_error: bool,
```
samdealy, line 17: As someone not familiar with this code, it's hard to know what "rel" means in this context. Please consider renaming to "relocation."
```suggestion
    ignore_relocation_error: bool,
```

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,275 @@
...
+    /// Whether to ignore .init sections in the enclave.
+    ignore_init_sec_error: bool,
```
samdealy, line 19: Similar to above, please consider renaming to "section."
```suggestion
    ignore_init_section_error: bool,
```

samdealy (approved):

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+        self
+    }
+
+    /// Generate the command to sign the given enclave object with the given
+    /// private key and write the resulting enclave to the given path. Note
+    /// that online signatures are inherently insecure.
+    pub fn sign(
+        &mut self,
+        unsigned_enclave: &Path,
+        config_path: &Path,
```
nick-mobilecoin, line 62: :wrench: Here and elsewhere in the file `config_path` the `path` suffix seems redundant with the type being `Path`.  If a suffix is desired I would suggest `config_file` to communicate it's a `file` and not a `dir`

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// Combine an unsigned enclave and signature into the output enclave, after
+    /// checking the signature.
+    pub fn catsig(
+        &mut self,
+        unsigned_enclave: &Path,
+        config_path: &Path,
+        public_key_pem: &Path,
+        gendata_output: &Path,
+        signature: &Path,
+        output_enclave: &Path,
```
nick-mobilecoin, line 133: :wrench: 
```suggestion
        unsigned_enclave: impl AsRef<Path>,
        config_path: impl AsRef<Path>,
        public_key_pem: impl AsRef<Path>,
        gendata_output: impl AsRef<Path>,
        signature: impl AsRef<Path>,
        output_enclave: impl AsRef<Path>,
```

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    /// Combine an unsigned enclave and signature into the output enclave, after
+    /// checking the signature.
+    pub fn catsig(
+        &mut self,
+        unsigned_enclave: &Path,
+        config_path: &Path,
+        public_key_pem: &Path,
+        gendata_output: &Path,
```
nick-mobilecoin, line 131: :upside_down_face: not sure on this.
The `gendata_output` helps associate the `gendata` of the 2 step signing process, but I think at the point someone is doing a 2 step signing they've had to figure out what `gendata` and `catsig` does.
Perhaps `signing_material`

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// Whether to ignore the presence of relocations in the enclave shared
+    /// object.
+    ignore_relocation_error: bool,
+    /// Whether to ignore .init sections in the enclave.
+    ignore_init_section_error: bool,
+    /// Whether to re-sign a previously signed enclave (default: false)
+    resign: bool,
+}
+
+impl Default for SgxSign {
```
nick-mobilecoin, line 24: :wrench: See comment in `From` trait, I think this should go away.

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+//! Builder wrapper around SgxSign.
+
+use std::{
+    path::{Path, PathBuf},
+    process::Command,
+};
+
+/// This structure wraps the execution of sgx_sign commands.
+#[derive(Clone, Debug)]
+pub struct SgxSign {
```
nick-mobilecoin, line 12: :thought_balloon: `sgx` is used in many places in this crate. I know the command line utility is called `sgx_sign` but I wonder if this struct should be named `EnclaveSigner` or similar.

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+impl From<PathBuf> for SgxSign {
+    fn from(sgx_sign_path: PathBuf) -> Self {
+        Self {
+            sgx_sign_path,
+            ignore_relocation_error: false,
+            ignore_init_section_error: false,
+            resign: false,
+        }
+    }
+}
```
nick-mobilecoin, line 203: :wrench: I think this should change not sure exactly how:
1. Instead of being the location of `sgx_sign` this should be the location of the `enclave` file being signed and a builder method to add the config
2. Remove this from and have a `new()` that takes an enclave and a config file.  The config file is optional according to SGX sign, but it looks like the methods are all requiring it.  `dump` is an exception here but I'm thinking `dump` doesn't belong in `SgxSign` and it should be in a separate struct.
3. Add builder methods to add the input enclave file and config file.

I'm leaning toward 1 to match `Edger8r`. I'm not opposed to rethinking `Edger8r`
My reasoning is that I'm thinking of inputs -> outputs. "What does the SgxSign" operate on?
```mermaid
graph LR
    A[enclave] --> B(sgx_sign)
    C[config] --> B
    B --> D[signed enclave]
```
The path to the signing binary doesn't seem needed, all of the building logic needs the path to the SGX_SDK so that already had to be set up by clients and setting it up per tool seem unnecessary at this time.

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+        let cmd = format!(
+            "{:?}",
+            SgxSign::default().sign(
+                Path::new("unsigned_enclave"),
+                Path::new("config_path"),
+                Path::new("private_key"),
+                Path::new("output_enclave"),
+            )
+        );
+        let expected_cmd = "\"/opt/intel/sgxsdk/bin/x64/sgx_sign\" \"sign\" \"-enclave\" \"unsigned_enclave\" \"-config\" \"config_path\" \"-key\" \"private_key\" \"-out\" \"output_enclave\"";
```
nick-mobilecoin, line 220: :exploding_head: This is an awesome way to test Command
I had written some tests for Command and I ended up doing some nasty trait logic so I could mock out Command, it was ugly and intrusive.
This idea of having a method (preferably private) that returns the command and comparing the string, then either not testing the final execution or leaving it to something like [assert_cmd](https://docs.rs/assert_cmd/latest/assert_cmd/) is so much cleaner.

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+        if self.ignore_init_section_error {
+            cmd.arg("-ignore-init-sec-error");
+        }
+
+        if self.resign {
+            cmd.arg("-resign");
+        }
+
+        cmd
+    }
```
nick-mobilecoin, line 190: :thought_balloon: See comment in the `From` I'm thinking this method should go away and come back in a struct, in a subsequent PR, whose only porpoise is to dump signed enclave contents

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// Generate the command to sign the given enclave object with the given
+    /// private key and write the resulting enclave to the given path. Note
+    /// that online signatures are inherently insecure.
+    pub fn sign(
+        &mut self,
+        unsigned_enclave: &Path,
+        config_path: &Path,
+        private_key: &Path,
+        output_enclave: &Path,
+    ) -> Command {
```
nick-mobilecoin, line 65: :wrench: I think IFF this suggestion is agreed on it should be a subsequent PR
Having this return a `Command` feels like it's leaking the abstraction. Clients will need to know what kind of output and return codes `sgx_sign` produces. I also think it's a bit misleading in that if I call a method named `sign` I expect it to sign something.
Suggest keeping the name but have this execute the `Command` and return something like a `Result<PathBuf>` with the signed enclave. I would move the current methods constant to a private method `build_sign_command()` so that it can be easily testable.

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_generate_sign_command() {
```
nick-mobilecoin, line 210: :upside_down_face: nit suggest omitting "test", other functions too.
```suggestion
    fn generate_sign_command() {
```

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+        let cmd = format!(
+            "{:?}",
+            SgxSign::default().sign(
+                Path::new("unsigned_enclave"),
+                Path::new("config_path"),
+                Path::new("private_key"),
+                Path::new("output_enclave"),
+            )
+        );
```
nick-mobilecoin, line 219: :thought_balloon: This is probably sufficient for this logic, but I would suggest in the future having:
- Two happy path tests with different values, though unlikely it helps ensure someone didn't do something like hardcode "private_key" as the name of the key file
-  One test per each builder method, `allow_relocations()` to ensure they are properly added  to the command line
- One test with all of the builder methods just to show they play nicely together. This one requires pragmatism with the current 3 methods it's ok to do all, but if there were 5 or more it quickly becomes noisy to run them all in one test, but multiple should still be run.

nick-mobilecoin (approved):
mainly suggestions for rustims and general thoughts on the direction the struct(s) might go

sdk-tools/src/sign.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    /// Generate the command to sign the given enclave object with the given
+    /// private key and write the resulting enclave to the given path. Note
+    /// that online signatures are inherently insecure.
+    pub fn sign(
+        &mut self,
+        unsigned_enclave: &Path,
+        config_path: &Path,
+        private_key: &Path,
+        output_enclave: &Path,
```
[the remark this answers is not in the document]
jcape: I think the problems you're mentioning with changing this to `impl AsRef<Path>` are actually significant.

Firstly, with just `impl AsRef<Path>` that means "take ownership of a type which can be referenced as a path." This is why you need to change callers to use `as_ref()`, which is super awkward. This will also entail a `memmove()` if you just give it a `PathBuf` If you wanted to do this, it should be `&impl AsRef<Path>`, because we don't actually want to take ownership here.

The second issue is, as noted, the single `P` forces everything to be the same type, so you'd need `P1`, `P2`, `P3` to be useful (this is what `impl AsRef<Path>` desugars to, AFAIK).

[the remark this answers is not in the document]
nick-mobilecoin: Looking at https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3ecd960e21aa6cdd8dc328cc216ad275 
It seems that `&impl AsRef<Path>` is more restrictive, it forces all callers to pass in a reference, which prevents
```rust
call_function(Path::new("stuff");
```
And the `impl AsRef<Path>` allows references as well as owning.  The `as_ref()` should only need to be done in the callee.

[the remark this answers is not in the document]
jcape: Hmm, OK, complaints withdrawn :-)

sdk-tools/src/sign.rs
jcape: The original purpose was to just wrap the execution of the `sgx_sign` binary, without being opinionated about what you're using it for.

I'm also fine to remove `Default` (and therefore the dependency on `core-build`) and have `From<impl AsRef<Path>>` be the only constructor.

sdk-tools/src/sign.rs
jcape: IIRC, the reason I did it this way was because `Command` has three different execute methods: `spawn() -> Result<Child>`, `status() -> Result<ExitStatus>`, and `output() -> Result<Output>`, providing different levels of control.

So if you want to access stdout/stderr, it's capture in `output()`, if you only care about pass/fail, `status().success()` is less to worry about.

sdk-tools/src/sign.rs
jcape: 💯

sdk-tools/src/sign.rs
jcape: I think `SgxSign` is fine, because the scope is just to wrap this tool. If we want to streamline a higher-level task, they should be their own things that call this.

sdk-tools/src/sign.rs
varsha888: will be included in https://github.com/mobilecoinfoundation/sgx/issues/264

sdk-tools/src/sign.rs
varsha888: Will be fixed in https://github.com/mobilecoinfoundation/sgx/issues/264

sdk-tools/src/sign.rs
varsha888: Will be fixed in https://github.com/mobilecoinfoundation/sgx/issues/264

sdk-tools/src/sign.rs
varsha888: Will be fixed in https://github.com/mobilecoinfoundation/sgx/issues/264

sdk-tools/src/sign.rs
varsha888: https://github.com/mobilecoinfoundation/sgx/issues/264

jcape (approved):

nick-mobilecoin (approved):

samdealy (approved):

nick-mobilecoin (approved):

PR #291, opened 2023-03-06
Add `BitAnd` to `mc-sgx-core-types::ReportData`

nick-mobilecoin:

### DIFF
--- a/core/types/src/report.rs
+++ b/core/types/src/report.rs
@@ -5,6 +5,7 @@ use crate::{
     config_id::ConfigId, impl_newtype_for_bytestruct, key_request::KeyId, new_type_accessors_impls,
     Attributes, ConfigSvn, CpuSvn, FfiError, IsvSvn, MiscellaneousSelect, MrEnclave, MrSigner,
 };
+use core::ops::BitAnd;
 use mc_sgx_core_sys_types::{
     sgx_isvext_prod_id_t, sgx_isvfamily_id_t, sgx_mac_t, sgx_prod_id_t, sgx_report_body_t,
     sgx_report_data_t, sgx_report_t, SGX_CONFIGID_SIZE, SGX_CPUSVN_SIZE, SGX_HASH_SIZE,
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* main
  * **PR #291** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/291" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/291?utm_source=stack-comment).

jcape (approved):

core/types/src/report.rs
```diff
@@ -33,6 +34,22 @@ impl_newtype_for_bytestruct! {
...
+/// There are times when only part of [`ReportData`] is of interest. [`BitAnd`]
+/// allows clients to mask off the parts of [`ReportData`] that are not of
+/// interest.
+impl BitAnd for ReportData {
+    type Output = Self;
+
+    fn bitand(self, rhs: Self) -> Self::Output {
+        // NB: Due to use in verification, this must be constant time.
+        let mut output = Self::default();
+        for (i, (a, b)) in self.0.d.iter().zip(rhs.0.d.iter()).enumerate() {
```
awygle, line 46: Wow, this is _dense_. I think I understand what's happening here but I think a simple `for i in range` type loop would be a lot more readable.

awygle (approved):
There's nothing wrong with this but I think the iterator magic is a bit opaque.

core/types/src/report.rs
nick-mobilecoin: Do you have a suggested implementation that you're thinking of?
Just for sanity I did this https://godbolt.org/z/W8ae5rPM8 to provide a bit more confidence the compiler won't care either way (if one were micro-optimizing), but also to show constant time ass.

I don't have a preference one way or the other. (though I have some scars from embedded compilers that would re-compute the array offset even in release :()
```rust
        for (i, o) in output.0.d.iter_mut().enumerate() {
            *o = self.0.d[i] & rhs.0.d[i];

        }
```

samdealy (approved):

core/types/src/report.rs
nick-mobilecoin: Going to merge for now and we can discuss how dense we want to get in iterators moving forward

PR #301, opened 2023-03-15
Add `Display` implementation for newtype wrappers

nick-mobilecoin:
Previously the newtype implementations didn't implement the `Display`
trait. Now the `Display` trait is implemented on any newtype that uses
one of `impl_newtype` or `impl_newtype_for_bytestruct`.

Renamed the internal macro `new_type_accessors_impls` to
`impl_newtype` for consistency.

### DIFF
--- a/core/types/src/macros.rs
+++ b/core/types/src/macros.rs
@@ -7,7 +7,7 @@ pub(crate) use alloc::vec::Vec;
 /// an SgxWrapperType that don't depend on the contents of the inner
 /// type.
 #[macro_export]
-macro_rules! new_type_accessors_impls {
+macro_rules! newtype_accessors_impls {
     ($($wrapper:ident, $inner:ty;)*) => {$(
         impl AsMut<$inner> for $wrapper {
             fn as_mut(&mut self) -> &mut $inner {
@@ -42,6 +42,23 @@ macro_rules! new_type_accessors_impls {
     )*}
 }
 
+/// Newtype wrapper for a primitive or struct type
+#[macro_export]
+macro_rules! impl_newtype {
+    ($($wrapper:ident, $inner:ty;)*) => {$(
+        $crate::newtype_accessors_impls! {
+            $wrapper, $inner;
+        }
+
+        impl core::fmt::Display for$wrapper {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                core::fmt::Debug::fmt(&self.0, f)
+            }
+        }
+
+    )*}
+}
+
 /// This macro provides common byte-handling operations when the type being
 /// wrapped is a struct containing a single fixed-size array of bytes.
 ///
@@ -50,7 +67,7 @@ macro_rules! new_type_accessors_impls {
 macro_rules! impl_newtype_for_bytestruct {
     ($($wrapper:ident, $inner:ident, $size:ident, $fieldname:ident;)*) => {$(
 
-        $crate::new_type_accessors_impls! {
+        $crate::newtype_accessors_impls! {
             $wrapper, $inner;
         }
 
@@ -100,12 +117,27 @@ macro_rules! impl_newtype_for_bytestruct {
             }
         }
 
+        impl core::fmt::Display for $wrapper {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                write!(f, "0x")?;
+                let inner: &[u8] = self.as_ref();
+                for byte in inner {
+                    write!(f, "{:02X}", byte)?;
+                }
+                Ok(())
+            }
+        }
+
     )*}
 }
 
 #[cfg(test)]
 mod test {
+    extern crate std;
+
     use crate::FfiError;
+    use std::format;
+    use std::string::ToString;
     use yare::parameterized;
 
     const FIELD_SIZE: usize = 24;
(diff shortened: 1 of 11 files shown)


nick-mobilecoin:
* main
  * **PR #301** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/301" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/301?utm_source=stack-comment).

core/types/src/macros.rs
```diff
@@ -42,6 +42,27 @@ macro_rules! new_type_accessors_impls {
...
+        $crate::newtype_accessors_impls! {
+            $wrapper, $inner;
+        }
+
+        impl core::fmt::Display for$wrapper {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                if f.alternate() {
+                    write!(f, "{:#?}", self.0)
+                } else {
+                    write!(f, "{:?}", self.0)
```
nick-mobilecoin, line 58: Not fond of re-directing to debug here, but it seemed like a nicer display than the struct display would be a case by case basis, and should be handled as they arise.

core/types/src/macros.rs
```diff
@@ -100,12 +121,27 @@ macro_rules! impl_newtype_for_bytestruct {
...
         }
 
+        impl core::fmt::Display for $wrapper {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                write!(f, "0x")?;
+                let inner: &[u8] = self.as_ref();
+                for byte in inner {
+                    write!(f, "{:02X}", byte)?;
+                }
+                Ok(())
```
nick-mobilecoin, line 131: It looks like https://doc.rust-lang.org/std/fmt/trait.UpperHex.html isn't implemented for &[u8] and it seemed a little heavy to bring in the hex crate for only this.

core/types/src/macros.rs
```diff
@@ -178,4 +214,67 @@ mod test {
...
+
+    #[test]
+    fn newtype_byte_array_display() {
+        let outer = Outer::from([
+            0xABu8, 0x00, 0xcd, 0x12, 0xfe, 0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x7, 0x8, 0x9, 0x0a,
+            0x0B, 0x0C, 0x0D, 0x0E, 0x0F, 0x10, 0x11, 0x12, 0x13,
+        ]);
+        assert_eq!(
+            outer.to_string(),
+            "0xAB00CD12FE0102030405060708090A0B0C0D0E0F10111213"
```
nick-mobilecoin, line 226: Note this implementation is upper hex, I don't have a preference. we could leverage alternate `#` for upper vs lower, but I don't think many people will think to look there to toggle case.

core/types/src/macros.rs
```diff
@@ -42,6 +42,23 @@ macro_rules! new_type_accessors_impls {
...
+#[macro_export]
+macro_rules! impl_newtype {
+    ($($wrapper:ident, $inner:ty;)*) => {$(
+        $crate::newtype_accessors_impls! {
+            $wrapper, $inner;
+        }
+
+        impl core::fmt::Display for$wrapper {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                core::fmt::Debug::fmt(&self.0, f)
```
nick-mobilecoin, line 55: I didn't like redirecting to `Debug` here but it felt like we would normally want the struct like appearances and it felt unnecessary to implement display for each C type.
Thinking IFF we need we can implement Display on a case by case bases for special C types.

jcape (changes requested):
I'm surprised the missing space before `$wrapper` compiles...

core/types/src/macros.rs
nick-mobilecoin: added underscores.
Also moved this to an `UpperHex` impl and have `Display` redirect to that.

core/types/src/macros.rs
```diff
@@ -42,6 +42,27 @@ macro_rules! new_type_accessors_impls {
...
 
+/// Newtype wrapper for a primitive or struct type
+#[macro_export]
+macro_rules! impl_newtype {
+    ($($wrapper:ident, $inner:ty;)*) => {$(
+        $crate::newtype_accessors_impls! {
+            $wrapper, $inner;
+        }
+
+        impl core::fmt::Display for$wrapper {
```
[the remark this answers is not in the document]
nick-mobilecoin: Fixed space and updated all `core` usages to absolute paths.

core/types/src/macros.rs
```diff
@@ -42,6 +42,23 @@ macro_rules! new_type_accessors_impls {
     )*}
 }
 
+/// Newtype wrapper for a primitive or struct type
+#[macro_export]
+macro_rules! impl_newtype {
```
varsha888, line 47: why is this not `new_type` ? 🤔

nick-mobilecoin: I'll update the description of the PR,
It looks like all the rust docs treat it as one word [newtype](https://doc.rust-lang.org/book/ch19-04-advanced-types.html?#using-the-newtype-pattern-for-type-safety-and-abstraction)

varsha888 (approved):

jcape (approved):

PR #302, opened 2023-03-24
ConstantTimeEq proc macro derive

### LINKED ISSUE #280: Impl ConstantTimeEq on quote types
Comment from https://github.com/mobilecoinfoundation/attestation/pull/3#discussion_r1116260392
> I'd say we should just impl ConstantTimeEq for Attributes in core-types, honestly.

varsha888:
### Motivation
Fixes #280 
Using [Subtle crate](https://crates.io/crates/subtle)
Implemented ConstantTimeEq using proc macro derive 
Implemented ConstantTimeGreater and ConstantTimeLesser

### Future Work
Move proc macro into subtle crate

### DIFF
--- a/constant_time_derive/Cargo.toml
+++ b/constant_time_derive/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "constant_time_derive"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1.0.8"
+quote = "1.0"
+subtle = { version = "2.4.0", default-features = false }
+syn = "1.0"
--- a/constant_time_derive/src/lib.rs
+++ b/constant_time_derive/src/lib.rs
@@ -1,0 +1,115 @@
+// Copyright (c) 2023 The MobileCoin Foundation
+
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Data, DataEnum, DeriveInput, Fields, GenericParam, Generics};
+
+#[proc_macro_derive(ConstantTimeEq)]
+pub fn constant_time_eq(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    derive_ct_eq(&input)
+}
+
+// TODO: Check or remove padding and align decorators on the struct
+fn parse_fields(fields: &Fields) -> Result<proc_macro2::TokenStream, &'static str> {
+    match &fields {
+        Fields::Named(fields_named) => {
+            let mut token_stream = quote!();
+            let mut iter = fields_named.named.iter().peekable();
+
+            while let Some(field) = iter.next() {
+                let ident = &field.ident;
+                match iter.peek() {
+                    None => token_stream.extend(quote! { {self.#ident}.ct_eq(&{other.#ident}) }),
+                    Some(_) => {
+                        token_stream.extend(quote! { {self.#ident}.ct_eq(&{other.#ident}) & })
+                    }
+                }
+            }
+            Ok(token_stream)
+        }
+        Fields::Unnamed(unnamed_fields) => {
+            let mut token_stream = quote!();
+            let mut iter = unnamed_fields.unnamed.iter().peekable();
+            let mut idx = 0;
+            while let Some(_) = iter.next() {
+                let i = syn::Index::from(idx);
+                match iter.peek() {
+                    None => token_stream.extend(quote! { {self.#i}.ct_eq(&{other.#i}) }),
+                    Some(_) => {
+                        token_stream.extend(quote! { {self.#i}.ct_eq(&{other.#i}) & });
+                        idx += 1;
+                    }
+                }
+            }
+
+            Ok(token_stream)
+        }
+        Fields::Unit => Err("Constant time cannot be derived for unit fields"),
+    }
+}
+
+fn parse_enum(data_enum: &DataEnum) -> Result<proc_macro2::TokenStream, &'static str> {
+    for variant in data_enum.variants.iter() {
+        if let Fields::Unnamed(_) = variant.fields {
+            panic!("Cannot derive ct_eq for fields in enums")
+        }
+    }
+    let token_stream = quote! {
+        ::subtle::Choice::from((self == other) as u8)
+    };
+
+    Ok(token_stream)
+}
+
+fn parse_data(data: &Data) -> Result<proc_macro2::TokenStream, &'static str> {
+    match data {
+        Data::Struct(variant_data) => parse_fields(&variant_data.fields),
+        Data::Enum(data_enum) => parse_enum(data_enum),
+        Data::Union(..) => Err("Constant time cannot be derived for a union"),
+    }
+}
+
+fn parse_lifetime(generics: &Generics) -> u32 {
+    let mut count = 0;
+    for i in generics.params.iter() {
+        if let GenericParam::Lifetime(_) = i {
+            count += 1;
+        }
+    }
+    count
+}
+
+fn derive_ct_eq(input: &DeriveInput) -> TokenStream {
+    let ident = &input.ident;
+    let data = &input.data;
+    let generics = &input.generics;
+    let is_lifetime = parse_lifetime(generics);
+    let ct_eq_stream: proc_macro2::TokenStream =
+        parse_data(data).expect("Failed to parse DeriveInput data");
+    let data_ident = if is_lifetime != 0 {
+        let mut s = format!("{}<'_", ident);
+
+        for _ in 1..is_lifetime {
+            s.push_str(", '_");
+        }
+        s.push('>');
+
+        s
+    } else {
+        ident.to_string()
+    };
+    let ident_stream: proc_macro2::TokenStream =
+        data_ident.parse().expect("Should be valid lifetime tokens");
+
+    let expanded: proc_macro2::TokenStream = quote! {
+            impl ::subtle::ConstantTimeEq for #ident_stream {
+                fn ct_eq(&self, other: &Self) -> ::subtle::Choice {
+                    use ::subtle::ConstantTimeEq;
+                    return #ct_eq_stream
+                }
+            }
+    };
+
+    expanded.into()
+}
--- a/core/build/src/lib.rs
+++ b/core/build/src/lib.rs
@@ -117,6 +117,8 @@ pub struct SgxParseCallbacks {
 
     // Dynamically Sized types
     dynamically_sized_types: Vec<String>,
+
+    constant_time_types: Vec<String>,
 }
 
 impl SgxParseCallbacks {
@@ -176,6 +178,20 @@ impl SgxParseCallbacks {
         self
     }
 
+    /// Types to derive constant time for
+    ///
+    /// # Arguments
+    /// * `constant_time_types` - Types to derive ConstantTimeEq for.
+    pub fn derive_constant_time<'a, E, I>(mut self, constant_time_types: I) -> Self
+    where
+        I: IntoIterator<Item = &'a E>,
+        E: ToString + 'a + ?Sized,
+    {
+        self.constant_time_types
+            .extend(constant_time_types.into_iter().map(ToString::to_string));
+        self
+    }
+
     /// Dynamically Sized Types
     ///
     /// # Arguments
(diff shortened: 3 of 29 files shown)


core/types/Cargo.toml
```diff
@@ -18,12 +18,14 @@ alloc = []
...
 [dependencies]
 bitflags = "2.0.0"
+constant_time_derive = { path = "../../constant_time_derive", version = "0.1.0" }
 displaydoc = { version = "0.2.3", default-features = false }
 mc-sgx-core-sys-types = { path = "../sys/types", version = "=0.5.1-beta.0" }
 mc-sgx-util = { path = "../../util", version = "=0.5.1-beta.0" }
 nom = { version = "7.1.2", default-features = false }
 rand_core = { version = "0.6.4", default-features = false }
 serde = { version = "1.0.152", default-features = false, features = ["derive"], optional = true }
+subtle = { version = "2.4.0", default-features = false, features = ["i128"] }
```
nick-mobilecoin, line 28: :question: I'm not sure we want to pull in i128, it's not available on all platforms and we want the types to work on HW wallets.
It may be that we want CT behind feature flags, but it still seems like it might be an attack vector if a malicious app times how long it takes a HW wallet to verify.

constant_time_derive/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+
+fn derive_ct_eq(input: &DeriveInput) -> TokenStream {
+    let ident = &input.ident;
+    let data = &input.data;
+    let generics = &input.generics;
+
+    let is_lifetime = parse_lifetime(generics);
+    let ct_eq_stream: proc_macro2::TokenStream =
+        parse_data(data).expect("Failed to parse DeriveInput data");
+    let data_ident = if is_lifetime {
```
NotGyro, line 82: Does this break on types with multiple explicit lifetimes?

core/build/src/lib.rs
```diff
@@ -176,6 +178,20 @@ impl SgxParseCallbacks {
         self
     }
 
+    /// Types to derive constant time for
+    ///
+    /// # Arguments
+    /// * `default_types` - Types to derive default for.
```
nick-mobilecoin, line 184: :wrench: Wrong docs, though one could probably omit the arguments from the docs as it's named clearly and I think the other comments were me just being pedantic. It looks like rust documentation suggest not documenting arguments if they are clearly communicated by name.
```suggestion
    /// * `constant_time_types` - Types to derive constant time for.
```

core/types/src/attestation_key.rs
```diff
@@ -251,6 +253,91 @@ mod test {
...
+    }
+
+    #[test]
+    fn ct_eq_version() {
+        let first = Version(2);
+        let second = Version(2);
+        let choice_result = first.ct_eq(&second);
+        let result: bool = From::from(choice_result);
+        assert!(result);
+    }
```
nick-mobilecoin, line 272: :thought_balloon: I'm not sure tests at this level are very beneficial:
1. These are only testing `==` nothing is testing for `!=`.
2. They feel pretty noisy here and don't provide confidence on the full logic of the macro itself

Suggest making a `tests` directory in the macro crate and test the different types there.
This will:
- make it easier to test boundary conditions there
- make the tests live with the crate if it gets moved somewhere else
- prevents noise here. I know that we're testing the to/from logic. but we're not testing hash, ord, etc here.
```rust
use subtle::ConstantTimeEq;
use constant_time_derive::ConstantTimeEq;

#[derive(ConstantTimeEq)]
struct NamedFields {
    a: u8,
    b: u8,
}

#[test]
fn named_fields() {
    let a = NamedFields { a: 1, b: 2 };
    let b = NamedFields { a: 1, b: 2 };
    let c = NamedFields { a: 1, b: 3 };
    assert!(bool::from(a.ct_eq(&b)));
    assert!(bool::from(!a.ct_eq(&c)));
}

```

constant_time_derive/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+                    }
+                }
+            }
+            Ok(token_stream)
+        }
+        Fields::Unnamed(_) => {
+            let token_stream = quote! { {self.0}.ct_eq({&other.0}) };
+
+            Ok(token_stream)
+        }
```
nick-mobilecoin, line 34: :wrench: TBH I'm not sure exactly what's going on here, but i think it only looks at the first field
```rust
#[derive(ConstantTimeEq)]
struct UnnamedFields(u8, u8);

#[test]
fn unnamed_fields() {
    let a = UnnamedFields(3, 4);
    let b = UnnamedFields(3, 4);
    let c = UnnamedFields(3, 3);
    assert!(bool::from(a.ct_eq(&b)));
    assert!(bool::from(!a.ct_eq(&c)));
}
```
That test fails with
```console
assertion failed: bool::from(!a.ct_eq(&c))
thread 'unnamed_fields' panicked at 'assertion failed: bool::from(!a.ct_eq(&c))', constant_time_derive/tests/newtype_struct.rs:29:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:64:14
   2: core::panicking::panic
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:111:5
   3: newtype_struct::unnamed_fields
             at ./tests/newtype_struct.rs:29:5
   4: newtype_struct::unnamed_fields::{{closure}}
             at ./tests/newtype_struct.rs:24:21
   5: core::ops::function::FnOnce::call_once
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/ops/function.rs:507:5
   6: core::ops::function::FnOnce::call_once
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/ops/function.rs:507:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

nick-mobilecoin (commented):
submitting without specific approval/denial since I'll be awol for the next couple of weeks

jcape (changes requested):

constant_time_derive/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "constant_time_derive"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
```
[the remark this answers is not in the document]
varsha888: It should be ok as proc macro for constantTimeEq will move into subtle.

core/types/Cargo.toml
varsha888: @jcape said we can change it if we come across a problem. @jcape thoughts?

core/build/src/lib.rs
varsha888: Good catch. Changed it

constant_time_derive/src/lib.rs
varsha888: I misunderstood that unnamed structs will be for the FFI types. Code should handle this now.

constant_time_derive/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+            panic!("Cannot derive ct_eq for fields in enums")
+        }
+    }
+    let token_stream = quote! {
+        if self == other {
+            ::subtle::Choice::from(1)
+        }
+        else {
+            ::subtle::Choice::from(0)
+        }
```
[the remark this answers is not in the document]
varsha888: Choice doesnt not have from impl for bool. It has to be added into subtle to do this. I can include it in my PR to subtle.

constant_time_derive/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+            while let Some(field) = iter.next() {
+                let ident = &field.ident;
+                match iter.peek() {
+                    None => token_stream.extend(quote! { {self.#ident}.ct_eq(&{other.#ident}) }),
+                    Some(_) => {
+                        token_stream.extend(quote! { {self.#ident}.ct_eq(&{other.#ident}) & })
+                    }
+                }
+            }
+            Ok(token_stream)
```
nick-mobilecoin, line 29: :clown_face: I don't know that this suggestion is much better. The while loop and peeking for "is there more" felt off
```suggestion
            let mut token_stream = quote!();
            let separators = ::core::iter::once("").chain(::core::iter::repeat("&"));
            let items = separators.zip(fields_named.named.iter());
            for (separator, field) in items {
                let ident = &field.ident;
                let separator_token: proc_macro2::TokenStream = separator.parse().expect("Should be a valid token");
                token_stream.extend(quote! { #separator_token {self.#ident}.ct_eq(&{other.#ident}) });
            }
            Ok(token_stream)
```

constant_time_derive/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+        for _ in 1..is_lifetime {
+            s.push_str(", '_");
+        }
+        s.push('>');
+
+        s
+    } else {
+        ident.to_string()
+    };
+    let ident_stream: proc_macro2::TokenStream = data_ident.parse().unwrap();
```
nick-mobilecoin, line 107: :upside_down_face: We've been trying to use expects over unwraps as the "reason" message will often play the same role as a comment, it also makes it easier to automatically check for no use of unwrap
```suggestion
    let ident_stream: proc_macro2::TokenStream =
        data_ident.parse().expect("Should be valid lifetime tokens");
```

core/types/src/quote.rs
```diff
@@ -394,4 +396,208 @@ mod test {
...
+
+    #[test]
+    fn ct_eq_quoting_enclave_report_info() {
+        let nonce = sgx_quote_nonce_t { rand: [1u8; 16] };
+        let app_enclave_target_info = sgx_target_info_t {
+            mr_enclave: Default::default(),
+            attributes: Default::default(),
+            reserved1: [1u8; 2],
+            config_svn: 0,
+            misc_select: 0,
```
nick-mobilecoin, line 408: :question: I think we can come back later and decide, but I wonder if the reserved fields should be excluded

core/types/src/svn.rs
```diff
@@ -2,41 +2,73 @@
...
 
 impl_newtype_for_bytestruct! {
     CpuSvn, sgx_cpu_svn_t, SGX_CPUSVN_SIZE, svn;
 }
 
+impl ConstantTimeGreater for CpuSvn {
+    fn ct_gt(&self, other: &Self) -> Choice {
+        if !self.0.svn.len().eq(&other.0.svn.len()) {
+            return Choice::from(0);
+        }
```
nick-mobilecoin, line 56: :wrench: This check isn't needed since the array is defined with a size inside of a struct
```suggestion
```

nick-mobilecoin (changes requested):
The main thing that should be done is mapping the `==` to Choice in the enum derive logic.
the other stuff is mainly nits and suggestions

constant_time_derive/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+        if let Fields::Unnamed(_) = variant.fields {
+            panic!("Cannot derive ct_eq for fields in enums")
+        }
+    }
+    let token_stream = quote! {
+        if self == other {
+            ::subtle::Choice::from(1)
+        }
+        else {
+            ::subtle::Choice::from(0)
```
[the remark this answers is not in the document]
varsha888: Interesting, Java or Kotlin didnt do this.

constant_time_derive/src/lib.rs
varsha888: 👍

core/types/src/quote.rs
varsha888: Why should it be excluded?

constant_time_derive/src/lib.rs
varsha888: I felt peeking is more readable but thanks for this, I might use this elsewhere

core/types/src/quote.rs
nick-mobilecoin: Often reserved fields are for future expansion(and memory alignment). Often times they're whatever happened to be on the stack for the structure.

nick-mobilecoin (approved):

nick-mobilecoin:
closing due to shifting focus

PR #308, opened 2023-03-31
Implement `Display` for`Attributes`

samdealy:
### Motivation
This PR implements `Display` for `Attributes` in order to make attestation errors more readable and useful (see [54](https://github.com/mobilecoinfoundation/attestation/pull/54)). Previously, we implemented `Display` in the `new_types!` macro and relied on the `Debug` trait. The new `Display` implementation prints out each `Attributes` flag in hex. 

### Future Work
- Implement `Display`for other `mc_sgx_core_types`

PR #310, opened 2023-04-03
Implement `Display` for `CpuSvn`

samdealy:
### Motivation
This PR implements `Display` for `CpuSvn` in order to make `attestation` repo errors more readable and useful (see https://github.com/mobilecoinfoundation/attestation/pull/54). Previously, we implemented `Display` in the `new_types!` macro and relied on the `Debug` trait. The new `Display` implementation prints out the `CpuSvn` byte array in `UpperHex` format.

### Future Work
- Continue implementing `Display` trait for `mc_sgx_core_types`

### DIFF
--- a/core/types/src/macros.rs
+++ b/core/types/src/macros.rs
@@ -101,7 +101,7 @@ macro_rules! impl_newtype_for_bytestruct {
             type Error = $crate::FfiError;
 
             fn try_from(src: &[u8]) -> ::core::result::Result<Self, Self::Error> {
-                if src.len() != $size {
+                if src.len() < $size {
                     return Err($crate::FfiError::InvalidInputLength);
                 }
 
--- a/core/types/src/macros.rs
+++ b/core/types/src/macros.rs
@@ -101,7 +101,7 @@ macro_rules! impl_newtype_for_bytestruct {
             type Error = $crate::FfiError;
 
             fn try_from(src: &[u8]) -> ::core::result::Result<Self, Self::Error> {
-                if src.len() < $size {
+                if src.len() != $size {
                     return Err($crate::FfiError::InvalidInputLength);
                 }
 
--- a/core/types/src/svn.rs
+++ b/core/types/src/svn.rs
@@ -30,3 +30,21 @@ pub struct CpuSvn(sgx_cpu_svn_t);
 impl_newtype_for_bytestruct! {
     CpuSvn, sgx_cpu_svn_t, SGX_CPUSVN_SIZE, svn;
 }
+
+#[cfg(test)]
+mod test {
+    extern crate std;
+
+    use super::*;
+    use std::format;
+
+    #[test]
+    fn cpu_svn_display() {
+        let cpu_svn = CpuSvn::from([1u8; CpuSvn::SIZE]);
+
+        let display_string = format!("{}", cpu_svn);
+        let expected_string = "0x0101_0101_0101_0101_0101_0101_0101_0101";
+
+        assert_eq!(display_string, expected_string);
+    }
+}


samdealy:
* main
  * **PR #308** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/308" target="_blank">Graphite</a> 
    * **PR #310** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/310" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/310?utm_source=stack-comment).

core/types/src/macros.rs
```diff
@@ -68,6 +68,83 @@ macro_rules! impl_newtype {
     )*}
 }
 
+/// This macro provides common byte-handling operations when the type being
+/// wrapped is a struct containing a single fixed-size array of bytes.
+///
+/// This should be called from within a private submodule.
+/// TODO: Delete this once Display has been implemented for all callers.
```
nick-mobilecoin, line 75: :wrench: I understand the desire is that this is meant to be a temp placeholder, but temp placeholders have a habit of persisting at times.
Consider updating `impl_newtype_for_bytestruct` to be something like:
```rust
#[macro_export]
macro_rules! impl_newtype_for_bytestruct {
    ($($wrapper:ident, $inner:ident, $size:ident, $fieldname:ident;)*) => {$(

        $crate::impl_newtype_for_bytestruct_no_display! {
            $wrapper, $inner, $size, $fieldname;
        }

        impl ::core::fmt::Display for $wrapper {
            fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
                write!(f, "{:#X}", self)
            }
        }

    )*}
}
```

core/types/src/svn.rs
```diff
@@ -27,6 +28,30 @@ impl_newtype! {
 #[derive(Default, Debug, Clone, Hash, PartialEq, Eq)]
 pub struct CpuSvn(sgx_cpu_svn_t);
 
-impl_newtype_for_bytestruct! {
+impl Display for CpuSvn {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        write!(f, "CpuSvn: {:X}", self)
```
nick-mobilecoin, line 33: :question: would it be better to update the display in `impl_newtype_for_bytestruct` to be of the form:
```rust
        impl ::core::fmt::Display for $wrapper {
            fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
                write!(f, "{}: ", stringify!($wrapper))?;
                ::core::fmt::UpperHex::fmt(self, f)
            }
        }
```
Note: used multiline write/fmt as this allows the passing of the alternate flag unconditionally via the `Formatter`, https://doc.rust-lang.org/std/fmt/trait.UpperHex.html

nick-mobilecoin (approved):
LGTM, a couple of thoughts/questions on the future roadmap

core/types/src/svn.rs
jcape: Display should not include the name of the struct.

jcape: Also, AFAIK, `UpperHex` will print `[0xDE, 0xAD, 0xBE, 0xEF]`, which isn't what we actually want, FWIW.

core/types/src/macros.rs
```diff
@@ -68,6 +68,83 @@ macro_rules! impl_newtype {
...
+            fn as_mut(&mut self) -> &mut [u8] {
+                &mut (self.0).$fieldname[..]
+            }
+        }
+
+        impl<'bytes> TryFrom<&'bytes [u8]> for $wrapper {
+            type Error = $crate::FfiError;
+
+            fn try_from(src: &[u8]) -> ::core::result::Result<Self, Self::Error> {
+                if src.len() < $size {
```
jcape, line 105: We should either use `Self::SIZE` or `$size`, but not both (note that there's also some usefulness to making `Self::SIZE` into a trait instead). Also, there's no reason to be lax in the size of the slice we accept, we're not returning anything useful to "advance" a buffer here.
```suggestion
                if src.len() != $size {
```

jcape (approved):

core/types/src/svn.rs
nick-mobilecoin: What's the impetus to move display out of the macro if it's going to be implemented the same as the macro was?
I can understand it for the other types where you want to show flags instead of a `u32` or perhaps a byte array instead of an integer.
Is it so that *all* the Display impls are explicit?

core/types/src/macros.rs
samdealy: While this is outside the scope of the PR now that this is unchanged,  I fixed it.

core/types/src/svn.rs
samdealy: When I started on this stack of PRs, I wasn't sure if a blanket macro implementation would make sense because I thought each `Display` might be slightly different. Now that all of them are done, I agree that they can all use the same macro impl.

core/types/src/macros.rs
samdealy: On second thought, introducing this unrelated change breaks a few tests in the PR stack. Let's address this in a follow up PR since it isn't germane to the `Display` impl.

jcape (approved):

nick-mobilecoin (approved):

PR #334, opened 2023-05-23
Prep for release 0.6.1

nick-mobilecoin:


.github/workflows/ci.yaml
```diff
@@ -26,72 +26,6 @@ jobs:
         with:
           globs: "**/*.md"
```
nick-mobilecoin, line 28: removed crev since it was failing due to, https://github.com/crev-dev/cargo-crev/issues/598 and it seems like there isn't a desire to pursue it at this time.

Cargo.toml
```diff
@@ -49,9 +49,7 @@ lto = true
 
 [workspace.metadata.release]
 shared-version = true
-dev-version-ext = "beta.0"
 consolidate-commits = true
-consolidate-pushes = true
 tag-name = "v{{version}}"
```
nick-mobilecoin, line 53: These got removed since they were for an older `cargo-release` version and are no longer supported

.github/workflows/ci.yaml
cbeck88: :+1:

cbeck88 (approved):
thank you

PR #335, opened 2023-05-25
Update `Attributes::set_*` methods to use explicit types

nick-mobilecoin:
Previously `Attributes:set_flags()` and
`Attributes::set_extended_features_mask()` used `u64` for the bits. This
provided no type safety or assurances that the correct bits were being
set. Now both methods use a dedicated bitflag type `AttributeFlags` and
`ExtendedFeatureRequestMask`.

### DIFF
--- a/core/types/src/attributes.rs
+++ b/core/types/src/attributes.rs
@@ -26,8 +26,8 @@ impl Attributes {
     /// # Arguments
     ///
     /// * `flags` - The flags to be set in the attributes
-    pub fn set_flags(mut self, flags: u64) -> Self {
-        self.0.flags = flags;
+    pub fn set_flags(mut self, flags: AttributeFlags) -> Self {
+        self.0.flags = flags.bits();
         self
     }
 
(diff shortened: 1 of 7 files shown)


core/types/src/attributes.rs
```diff
@@ -36,15 +36,15 @@ impl Attributes {
...
+    pub fn set_extended_features_mask(mut self, features_mask: ExtendedFeatureRequestMask) -> Self {
+        self.0.xfrm = features_mask.bits();
         self
     }
 }
 
 impl Display for Attributes {
     fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
-        match Flags::from_bits(self.0.flags) {
+        match AttributeFlags::from_bits(self.0.flags) {
```
nick-mobilecoin, line 47: `Flags`, with no qualifier coming out of `mc-sgx-core-types` felt ambiguous so added `Attribute` prefix

core/types/src/attributes.rs
```diff
@@ -54,7 +54,7 @@ impl Display for Attributes {
             }
             None => write!(f, "Flags: (none)")?,
         }
-        match Xfrm::from_bits(self.0.xfrm) {
+        match ExtendedFeatureRequestMask::from_bits(self.0.xfrm) {
```
nick-mobilecoin, line 57: I'm fine if people think this should go back to `Xfrm` to match Intel's usage.
My background has me see `Xfrm` as "transform" and doing a quick google search for xfrm IP transforms seem to come up the most.

eranrund: I like the explicit name, being new to looking at a lot of SGX terminology and all the obscure acronyms involved.

eranrund (approved):

cbeck88 (approved):

PR #342, opened 2023-06-13
Add `mc-sgx-dcap-types::Collateral`

nick-mobilecoin:
Add a more ergonomic rust collateral type hide the details of the
`sgx_ql_qve_collateral_t` type.


dcap/quoteverify/src/collateral.rs
```diff
@@ -58,6 +77,19 @@ impl<T: AsRef<[u8]>> Collateral for Quote3<T> {
         }
         .into_result()?;
 
+        // It shouldn't happen that `tee_qv_get_collateral()` returns success
+        // while keeping the collateral as a null pointer, but we defend against
+        // it since we don't control the C implementation.
+        if sgx_collateral.is_null() {
+            return Err(Error::QuoteLibrary(QlError::NoQuoteCollateralData));
```
nick-mobilecoin, line 84: :shrug: A bit undecided, could make a dedicated error but this error seemed good enough and I don't think this case is likely to be hit in practice as we probably got an error from `tee_qv_get_collateral()`

dcap/quoteverify/src/lib.rs
```diff
@@ -7,13 +7,13 @@ mod collateral;
...
-pub use collateral::{Collateral, CollateralPointer};
-use mc_sgx_dcap_types::QlError;
+pub use collateral::Collateral;
+use mc_sgx_dcap_types::{CollateralError, QlError};
 pub use quote_enclave::{LoadPolicyInitializer, PathInitializer};
 pub use verify::supplemental_data_size;
 
 /// Errors interacting with quote verification library functions
-#[derive(Clone, Debug, displaydoc::Display, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[derive(Clone, Debug, displaydoc::Display, Eq, PartialEq)]
```
nick-mobilecoin, line 16: Underlying errors don't implement hash or ord so removed them here

dcap/types/src/collateral.rs
```diff
@@ -0,0 +1,805 @@
...
+//      internally.  Currently, only values of 3.0 and 3.1 are valid.  Note, if you set this to 3.1, the PCCS use to retrieve
+//      verification collateral must support the new 3.1 APIs.
+//
+// This version can be ensured by setting the `pccs_api_version` key of `/etc/sgx_default_qcnl.conf`
+// to `3.1`.
+//
+//      ,"pccs_api_version": "3.1"
+//
+const VERSION_MAJOR: u16 = 3;
+const VERSION_MINOR: u16 = 1;
```
nick-mobilecoin, line 37: Since we have some sway on the enclave host systems I didn't see much need to support the other versions at this time.  If we do want to suggest a follow on PR as this one is already a bit large

dcap/types/src/collateral.rs
```diff
@@ -0,0 +1,805 @@
...
+    /// Error converting from DER {0}
+    Der(x509_cert::der::Error),
+    /// Unsupported version, expected 3.1, but got {0}.{1}
+    Version(u16, u16),
+    /// Unsupported collateral service, expected SGX (0), but got {0}
+    CollateralService(u32),
+    /// Error converting bytes to String {0}
+    Utf8(FromUtf8Error),
+    /// One of the collateral fields is missing
+    MissingCollateral,
```
nick-mobilecoin, line 51: This error feels a bit like a catch all, I wanted to keep the unsafe C buffer logic as isolated as possible.  One thought was to have `MissingCollateral(String)` and the C buffer function would get passed in the string "tcb_info" or similar for the collateral it's processing.

At the same time, I'm not sure how likely we're to get this far with bad collateral data from sgx

dcap/types/src/collateral.rs
```diff
@@ -0,0 +1,805 @@
...
+/// - <https://api.portal.trustedservices.intel.com/documentation#pcs-tcb-info-v4>
+/// - <https://api.portal.trustedservices.intel.com/documentation#pcs-enclave-identity-v4>
+/// - <https://certificates.trustedservices.intel.com/IntelSGXRootCA.der>
+///
+/// or the data can be retrieved from `sgx_ql_get_quote_verification_collateral()`
+///
+/// The certificate chains and CRLs are documented in
+/// <https://api.trustedservices.intel.com/documents/Intel_SGX_PCK_Certificate_CRL_Spec-1.5.pdf>
+#[derive(Debug, Clone)]
+pub struct Collateral {
```
nick-mobilecoin, line 79: Up for suggestions here, it feels like a lot of members.
we could make some more nested types if desired, like "tcb{issuer_chain, info}", "qe{issuer_chain, identity}", but not sure it buys us much

dcap/types/src/collateral.rs
```diff
@@ -0,0 +1,805 @@
...
+    root_ca_crl: CertificateList,
+    pck_crl_issuer_chain: Vec<Certificate>,
+    pck_crl: CertificateList,
+    tcb_issuer_chain: Vec<Certificate>,
+    tcb_info: String,
```
nick-mobilecoin, line 84: This String is what would be fed into https://github.com/mobilecoinfoundation/attestation/blob/945015c9cf6510766fddf8ecfe383cd5e29fc161/verifier/src/tcb.rs
I think perhaps we should consider moving that module into here and this should be a TCB type, which propagates an error out if the underlying str is invalid.

dcap/types/src/collateral.rs
```diff
@@ -0,0 +1,805 @@
...
+//      0x00000000: SGX or 0x00000081: TDX
+const SGX_TEE: u32 = 0;
+
+// From the comment in `/etc/sgx_default_qcnl.conf`:
+//
+//      If you use a PCCS service to get the quote verification collateral, you can specify which PCCS API version is to be used.
+//      The legacy 3.0 API will return CRLs in HEX encoded DER format and the sgx_ql_qve_collateral_t.version will be set to 3.0, while
+//      the new 3.1 API will return raw DER format and the sgx_ql_qve_collateral_t.version will be set to 3.1. The pccs_api_version
+//      setting is ignored if collateral_service is set to the Intel PCS. In this case, the pccs_api_version is forced to be 3.1
+//      internally.  Currently, only values of 3.0 and 3.1 are valid.  Note, if you set this to 3.1, the PCCS use to retrieve
```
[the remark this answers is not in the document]
nick-mobilecoin: fixed

dcap/types/src/collateral.rs
```diff
@@ -0,0 +1,805 @@
...
+    // valid. This function returns a value which copies from the `bytes` and
+    // thus no longer references the `bytes`.
+    #[allow(unsafe_code)]
+    let slice = unsafe { core::slice::from_raw_parts(bytes as *const u8, size as usize) };
+
+    // The DER encoding has an extra NULL byte at the end.
+    // Since DER is a binary format, there is nothing stopping the last valid
+    // byte from being 0, because of that we only trim the extra byte and not
+    // *all* trailing null bytes via the `trim_null_end()` function
+    let [crl @ .., _] = slice else {
```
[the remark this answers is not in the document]
nick-mobilecoin: updated to error out if the last byte is not null

eranrund (approved):

PR #350, opened 2023-06-22
Add `BitAnd` for `Attributes`

nick-mobilecoin:
The `Attributes` value of report body ng enclaves) needs to be masked as
part of the verifying of a report body. Add the `BitAnd` trait to the
`Attributes` type to make this operation easier.

### DIFF
--- a/core/types/src/attributes.rs
+++ b/core/types/src/attributes.rs
@@ -53,7 +53,11 @@ impl Display for Attributes {
                     write!(f, "Flags: {}", flags)?
                 }
             }
-            None => write!(f, "Flags: (none)")?,
+
+            None => {
+                write!(f, "Flags: ")?;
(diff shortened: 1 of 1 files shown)


core/types/src/attributes.rs
```diff
@@ -53,7 +53,11 @@ impl Display for Attributes {
                     write!(f, "Flags: {}", flags)?
                 }
             }
-            None => write!(f, "Flags: (none)")?,
+
```
nick-mobilecoin, line 56: When using the masking of all the bits `0xFFFF_FFFF_FFFF_FFFF` it seemed wrong to be outputting `(none)` here, figured if any error happens a display of the hex value would be better

eranrund (approved):

PR #382, opened 2023-07-27
Add `Serialize` and `Deserialize` traits to `Collateral`

nick-mobilecoin:
The `Collateral` type is used to go to/from an enclave from the
untrusted application, so it needs to support serialization.

### DIFF
--- a/dcap/types/src/collateral.rs
+++ b/dcap/types/src/collateral.rs
@@ -10,6 +10,7 @@ extern crate alloc;
 use alloc::string::{FromUtf8Error, String};
 use alloc::vec::Vec;
 use mc_sgx_dcap_sys_types::sgx_ql_qve_collateral_t;
+use serde::{Deserialize, Serialize};
 use x509_cert::crl::CertificateList;
 use x509_cert::der::Decode;
 use x509_cert::Certificate;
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
* main
  * **PR #382** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/382" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/382?utm_source=stack-comment).

dcap/types/src/collateral.rs
```diff
@@ -75,13 +76,18 @@ impl From<FromUtf8Error> for Error {
 ///
 /// The certificate chains and CRLs are documented in
 /// <https://api.trustedservices.intel.com/documents/Intel_SGX_PCK_Certificate_CRL_Spec-1.5.pdf>
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct Collateral {
+    #[serde(with = "certificate_list")]
```
nick-mobilecoin, line 81: I'm not fond of how manual this feels for serde, but the `X509_cert` types didn't derive serde, we could wrap them in NewTypes but we'd still be stuck manually implementing the de/serialize.  IFF we wanted to keep these as vecs of DER bytes on the Collateral we could get it for free, but then we'd need to convert to/from the Certificate and CertificateList types which is fallible.

eranrund (approved):

PR #397, opened 2023-08-08
(cargo-release) version 0.7.4

nick-mobilecoin:

PR #398, opened 2023-08-08
Fix `alloc` feature in `mc-sgx-dcap-types` missing dependency

nick-mobilecoin:
Previously the `alloc` feature in `mc-sgx-dcap-types` was not specifying
the need for `serde/alloc`. `serde/alloc` is now a dependency of the
`alloc` feature.

### DIFF
--- a/dcap/types/Cargo.toml
+++ b/dcap/types/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.62.1"
 
 [features]
 default = []
-alloc = ["dep:x509-cert"]
+alloc = ["dep:x509-cert", "serde/alloc"]
 tcb = ["alloc", "dep:x509-cert", "dep:const-oid", "dep:hex", "serde/alloc"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


nick-mobilecoin:
* main
  * **PR #397** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/397" target="_blank">Graphite</a> 
    * **PR #398** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/398" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/398?utm_source=stack-comment).

dcap/types/Cargo.toml
```diff
@@ -13,7 +13,7 @@ rust-version = "1.62.1"
 
 [features]
 default = []
-alloc = ["dep:x509-cert"]
+alloc = ["dep:x509-cert", "serde/alloc"]
```
nick-mobilecoin, line 16: Looks like I missed this in #393.
not fond that `alloc` isn't tested alone. I need to see what cargo release does to find this, as cargo package doesn't seem to notice the lack of dependency when building.

If this continues to be a problem it may be worth looking into something like https://github.com/frewsxcv/cargo-all-features

eranrund (approved):

PR #427, opened 2023-10-11
Trim null bytes from Collateral strings

nick-mobilecoin:
The `Collateral` strings are coming from a C api which includes the null
trailing byte in the length. In order to work with strict JSON parsers
the null byte is omitted from the final rust `String`.

### DIFF
--- a/dcap/types/src/collateral.rs
+++ b/dcap/types/src/collateral.rs
@@ -199,7 +199,18 @@ fn string_from_bytes(bytes: *mut core::ffi::c_char, size: u32) -> Result<String,
     // thus no longer references the `bytes`.
     #[allow(unsafe_code)]
     let slice = unsafe { core::slice::from_raw_parts(bytes as *const u8, size as usize) };
-    Ok(String::from_utf8(slice.to_vec())?)
+
+    let string = String::from_utf8(slice.to_vec())?;
+
+    // The C API returns a null terminated string, so we need to trim the null
+    // byte(s) off for Json parsers.
+    let trimmed = string.trim_matches('\0');
+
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
* main
  * **PR #427** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/427" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/427?utm_source=stack-comment).

dcap/types/src/collateral.rs
```diff
@@ -199,7 +199,18 @@ fn string_from_bytes(bytes: *mut core::ffi::c_char, size: u32) -> Result<String,
     // thus no longer references the `bytes`.
     #[allow(unsafe_code)]
     let slice = unsafe { core::slice::from_raw_parts(bytes as *const u8, size as usize) };
-    Ok(String::from_utf8(slice.to_vec())?)
+
+    let string = String::from_utf8(slice.to_vec())?;
+
+    // The C API returns a null terminated string, so we need to trim the null
+    // byte(s) off for Json parsers.
+    let trimmed = string.trim_matches('\0');
```
nick-mobilecoin, line 207: The null byte doesn't count as whitespace so `trim()` doesn't work here.

nick-mobilecoin:
That clippy (beta) build is non blocking.  I'm a bit torn on the beta builds, this isn't the first time it's cropped up because downstream is chaning, at the same time if we don't keep up on it then `stable` will start to fail...

PR #430, opened 2023-10-13
Update SGX SDK to 2.20.100.4

nick-mobilecoin:


nick-mobilecoin:
* main
  * **PR #430** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/430" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/430?utm_source=stack-comment).

.devcontainer/Dockerfile
```diff
@@ -12,14 +12,14 @@ RUN  apt-get update \
   && rm -r /var/lib/apt/lists
 
 # Install SGX_SDK
-ARG SGX_URL=https://download.01.org/intel-sgx/sgx-linux/2.19/distro/ubuntu22.04-server/sgx_linux_x64_sdk_2.19.100.3.bin
+ARG SGX_URL=https://download.01.org/intel-sgx/sgx-linux/2.20/distro/ubuntu22.04-server/sgx_linux_x64_sdk_2.20.100.4.bin
```
nick-mobilecoin, line 15: 2.21 is out, but I only bumped to 2.20 here to reduce the overall change size and keep the updates in separate prs.

core/build/headers/sgx_attributes.h
```diff
@@ -40,7 +40,9 @@
 #define SGX_FLAGS_MODE64BIT      0x0000000000000004ULL     /* If set, then the enclave is 64 bit */
 #define SGX_FLAGS_PROVISION_KEY  0x0000000000000010ULL     /* If set, then the enclave has access to provision key */
 #define SGX_FLAGS_EINITTOKEN_KEY 0x0000000000000020ULL     /* If set, then the enclave has access to EINITTOKEN key */
-#define SGX_FLAGS_KSS            0x0000000000000080ULL     /* If set enclave uses KSS */
+#define SGX_FLAGS_KSS            0x0000000000000080ULL     /* If set, then the enclave uses KSS */
```
nick-mobilecoin, line 43: These vendored header file updates come from, https://github.com/intel/linux-sgx/compare/sgx_2.19...sgx_2.20

Unfortunately the github web interface isn't very good for looking at the diff. I diffed the vendored headers locally with beyondcompare.

tcrypto/sys/src/lib.rs
```diff
@@ -21,41 +21,3 @@ pub use mc_sgx_tcrypto_sys_types::{
 };
 
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
```
nick-mobilecoin, line 23: See https://github.com/intel/linux-sgx/issues/958 this is a function meant to be used from the trusted side.
We don't currently have a way to unit test the trusted side of things, we were getting by calling into this from outside of an enclave, but don't think it's worth the extra work to pull in more C/C++ dependencies to be able to test the C bindings
This was one of the few places that we were testing the bindings to a trusted interface.

eranrund (approved):

PR #455, opened 2023-12-11
Fix reproducible builds due to how derives were populated

nick-mobilecoin:
Previously `HashSet` was used to prevent duplicate entries in the
computed derive attributes for bindgen generated code. The default
hasher used in `HashSet` is randomly seeded on each process run to
prevent HashDos. This random seeding resulted in changing the ordering
of the resulted derive attributes. This changing order of the derive
attributes prevented reproducible builds as the derived implementations
would move around in the resultant binary. Now `BtreeSet` is used to
prevent duplicate derives during code generation. This provides a
consistent ordering of the derives.

### DIFF
--- a/core/build/src/lib.rs
+++ b/core/build/src/lib.rs
@@ -7,7 +7,7 @@ use bindgen::{
     callbacks::{IntKind, ParseCallbacks},
     Builder, EnumVariation,
 };
-use std::collections::HashSet;
+use std::collections::BTreeSet;
 use std::{env, path::PathBuf};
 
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
* main
  * **PR #455** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/455?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/455?utm_source=stack-comment).

eranrund (approved):

PR #461, opened 2024-01-02
Use `lockfile-only` for cargo dependabot

nick-mobilecoin:
Previously the dependabot for cargo was using `auto` which would update
both the manifest, `Cargo.toml` files and the `Cargo.lock` file. Since
this repository is for library crates, we only want to update the lock
file that people develop with. We don't want to update the manifest file
and force consumers to update unnecessarily.

See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy in particularly notice that cargo only supports `auto, lockfile-only`. Ideally we'd use `increase-if-necessary`. An alternative may be to look into rennovate bot, it appears it may support a better option, https://docs.renovatebot.com/configuration-options/#rangestrategy

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -7,6 +7,7 @@ updates:
   open-pull-requests-limit: 25
   commit-message:
     prefix: "chore(deps)"
+  versioning-strategy: lockfile-only
 
 - package-ecosystem: github-actions
   directory: "/"


nick-mobilecoin:
* `main`
  * **PR #461** <a href="https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/461?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/mobilecoinfoundation/sgx/461?utm_source=stack-comment).

eranrund (approved):

PR #489, opened 2024-03-20
Paramerterize shared library values

dbrowne:
add the following constants 
QUOTING_ENCLAVE
PROVISIONING_CERTIFICATE_ENCLAVE
ID_ENCLAVE

### Motivation

Library names should not be hardcoded. The paths are not universal across Linux distributions.

### DIFF
--- a/dcap/ql/src/quote_enclave.rs
+++ b/dcap/ql/src/quote_enclave.rs
@@ -40,6 +40,13 @@ pub struct PathInitializer;
 
 static PATH_INITIALIZER: Lazy<Mutex<Option<PathInitializer>>> = Lazy::new(|| Mutex::new(None));
 
+const QUOTING_ENCLAVE: &str = "/usr/lib/x86_64-linux-gnu/libsgx_qe3.signed.so.1";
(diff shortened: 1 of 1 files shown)


dcap/ql/src/quote_enclave.rs
```diff
@@ -40,6 +40,13 @@ pub struct PathInitializer;
 
 static PATH_INITIALIZER: Lazy<Mutex<Option<PathInitializer>>> = Lazy::new(|| Mutex::new(None));
 
+const QUOTING_ENCLAVE: &str = "/usr/lib/x86_64-linux-gnu/libsgx_qe3.signed.so.1";
```
nick-mobilecoin, line 43: I read the following from the motivation
> Library names should not be hardcoded. The paths are not universal across Linux distributions.

but I'm not quite seeing how this address that. The paths are still hardcoded and aren't handling the different linux distributions.

I'm not advocating for logic to handle the different linux distributions here either. As it currently stands, it may be difficult to find the time to review a change such as that.

